PR #28, opened 2015-11-28
Fix OverflowException when decompressing DXT5 during bit shifting due…

babelshift:
… to integer overflow.

Unfortunately, whatever color format these DXT5 textures use causes them to look strange when decompiled. See below. This issue has been reported by others using tools like ModKit and s2dc, so I don't know if it's specific to whatever decompressing algorithm this is using.
loadingscreen_tga

I don't know enough about it to fix the issue, but the overflow should be fixed at least.

### DIFF
--- a/ValveResourceFormat/ThirdParty/DDSImage.cs
+++ b/ValveResourceFormat/ThirdParty/DDSImage.cs
@@ -141,7 +141,12 @@ namespace ValveResourceFormat.ThirdParty
             byte alpha1 = blockStorage[1];
 
             int bitOffset = 2;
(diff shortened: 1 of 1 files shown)


xPaw:
Perhaps the colors are in different byte order? Try playing with that.

There was a similar issue fixed in f14f3fffb57d86ea74738f1524703bf00be61a23

babelshift:
Looked and done. I'll try rearranging the byte ordering.

babelshift:
After some testing and research, I've found that the vtex_c files which are incorrectly colored on extraction are TGA files compressed with DXT5 and possibly in YCoCg colorspace.

Looking at the REDI tab reveals:
```
CResourceString m_String = "Texture Compiler Version Image YCoCg Conversion"
```

This is a bit out of my league. I'm not sure if Valve's texture compiler is converting to YCoCg colorspace prior to compression. Maybe that means it needs to be read out as YCoCg and converted to RGB during decompression?

xPaw:
Create a separate issue for with an example file, I will take a look later.

PR #31, opened 2015-11-30
Add search for vpk

babelshift:
Intended for issue: https://github.com/SteamDatabase/ValveResourceFormat/issues/25

This is a rather big commit, so it needs testing and feedback. Let me know what you guys think on the UI and UX. I'm certainly open to suggestions and improvements. Here's what I've added.

#### Search Form
Added a toolbar which currently only contains a Search button. I can imagine adding more buttons in future commits to the toolbar such as Open, Extract, and Properties.

The button is only enabled when a VPK file is open and currently selected/viewed. When enabled and clicked, a Search form is displayed with a textbox to enter search parameters, a combobox of the search type to perform, and whether or not it's a case sensitive search.

**Currently not implemented:**
  * wild card searches
  * case sensitive option

#### BetterTreeView Control
Inherits from TreeView but allows for performing a breadth-first-search against the TreeNodes in the TreeView. A collection of matching TreeNodes is returned. Matching is based on a search type such as "FileNameExact", "FileNamePartial", and "FullPath".

#### TreeViewWithSearchResult Control
Contains a BetterTreeView and a ListView. When a search is performed against the BetterTreeView, the results of the search are displayed in the ListView with file name, file size, and file type. Currently, it relies on the MainForm to orchestrate the whole thing, so it's a bit less than optimal at the moment.

Right clicking a ListViewItem will give the Extract context menu option.

Double clicking a ListViewItem will open that file.

**Watch out for:**
  * haven't thoroughly tested multiple VPK files open at once with searching (the intention is for search to find the TreeView of the SelectedTab and use that)

**Currently not implemented:** 
  * sorting when clicking column headers

**Room for improvement:** 
  * don't jump to TreeNode when right clicking on ListViewItem

#### General Notes ####
Some code got moved around for convenience (like building the TreeView from Package).

### DIFF
--- a/GUI/Controls/TreeViewWithSearchResults.cs
+++ b/GUI/Controls/TreeViewWithSearchResults.cs
@@ -1,0 +1,250 @@
+﻿using GUI.Forms;
+using System;
+using System.Windows.Forms;
+using ValveResourceFormat;
+using GUI.Utils;
+using System.Drawing;
+
+namespace GUI.Controls
+{
+    /// <summary>
+    /// Represents a user control in which a TreeView and ListView are used to view a directory/file listing. In addition to a normal TreeView,
+    /// this control allows for searching to occur within the TreeView and have the results displayed in a ListView with details about the resulting
+    /// items.
+    /// </summary>
+    public partial class TreeViewWithSearchResults : UserControl
+    {
+        private ImageList imageList;
+
+        public TreeNodeMouseClickEventHandler TreeNodeMouseDoubleClick;             // when a TreeNode is double clicked
+        public TreeNodeMouseClickEventHandler TreeNodeMouseClick;                   // when a TreeNode is single clicked
+        public EventHandler<ListViewItemClickEventArgs> ListViewItemDoubleClick;    // when a ListViewItem is double clicked
+        public EventHandler<ListViewItemClickEventArgs> ListViewItemRightClick;     // when a ListViewItem is single clicked
+
+        /// <summary>
+        /// Constructor to require an image list for display on listed TreeView nodes and ListView items.
+        /// </summary>
+        /// <param name="imageList"></param>
+        public TreeViewWithSearchResults(ImageList imageList) : this()
+        {
+            this.imageList = imageList;
+        }
+
+        /// <summary>
+        /// Require a default constructor for the designer.
+        /// </summary>
+        public TreeViewWithSearchResults()
+        {
+            InitializeComponent();
+
+            mainListView.MouseDoubleClick += MainListView_MouseDoubleClick;
+            mainListView.MouseDown += MainListView_MouseDown;
+            mainListView.Resize += MainListView_Resize;
+            mainListView.FullRowSelect = true;
+
+            mainTreeView.HideSelection = false;
+
+            mainTreeView.NodeMouseDoubleClick += (o, e) =>
+            {
+                if (TreeNodeMouseDoubleClick != null) { TreeNodeMouseDoubleClick(o, e); }
+            };
+
+            mainTreeView.NodeMouseClick += (o, e) =>
+            {
+                if (TreeNodeMouseClick != null)
+                {
+                    // if user left clicked a folder, show the contents of that folder in the list view
+                    if (e.Button == MouseButtons.Left && e.Node.Tag is TreeViewFolder)
+                    {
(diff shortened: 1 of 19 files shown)


babelshift:
When search form is open:
searchform

When search results are displayed:
searchresults

xPaw:
Thanks for doing this! @SinZ163 or I will test it some time to see if we find any other issues before merging.

xPaw:
Here are some things:
- Clicking on a folder in search result should focus on that folder in the tree
- Clicking on a folder in the tree should update the search results with items in that folder (something like GCFScape does)
- Display size for folders as count of items in that folder
- Make columns automatically scale to window size
- Make partial match the default search option
- Search options should be remembered (click case sens, search something, open search menu again, case sens should be still selected)

Basically looking at GCFScape should be a good inspiration, but this is a good start, thanks for doing it!

SinZ163:
Also there are now branch conflicts that need to be fixed before it can be merged easily now.

babelshift:
That's where I originally had it and then moved it, but I think I agree on second thought. Here's an image of what this latest commit looks like in action.

searchresultsfilesize

babelshift:
@SinZ163 merge conflicts should be resolved now.

babelshift:
Previous commit addressed the following:
  * Display size for folders as count of items in that folder
  * Make columns automatically scale to window size
  * Make partial match the default search option
  * Search options should be remembered (click case sens, search something, open search menu again, case sens should be still selected)
  * Case sensitive search

xPaw:
There are still some points I'd like to be fixed before I merge this:

- Clicking on a folder in search result should focus on that folder in the tree
- Clicking on a folder in the tree should update the search results with items in that folder (something like GCFScape does)

This will basically make the side list more useful, and not just waste space when you're not searching.

Also there's something wrong about displayed size, sometimes it says `0 B` even though the file is not empty and has some data.

babelshift:
I'll have some time to finish the rest of the items today and tomorrow.

babelshift:
Previous commit addresses the following:
  * Clicking on a folder in search result should focus on that folder in the tree. **Note that due to the way Windows handles focus, the highlight may be grayed out because the TreeView doesn't have focus and I couldn't find a reliable way to grab focus without causing some weird glitches. So for now its highlighted but may be hard to see. Regardless, the node is jumped to even if you scroll the node out of view.**
  * Clicking on a folder in the tree should update the search results with items in that folder (something like GCFScape does)
  * Double clicking a folder in the search results now displays its contents in the search results

GUI/Controls/TreeViewWithSearchResults.cs
```diff
@@ -0,0 +1,250 @@
...
+        private void AddNodeToListView(TreeNode node)
+        {
+            ListViewItem item = new ListViewItem(node.Text);
+            item.ImageKey = node.ImageKey;
+            item.Tag = node;
+
+            if (node.Tag.GetType() == typeof(PackageEntry))
+            {
+                var file = node.Tag as PackageEntry;
+                item.SubItems.Add(file.Length.ToFileSizeString());
```
xPaw: This should be `(file.SmallData.Length + file.Length)` to display correct size.

xPaw:
This is better now. If you could squash your commits before I merge, that would be good.

GUI/Controls/TreeViewWithSearchResults.cs
babelshift: @xPaw What do you think about adding a property like `FullLength` or `TotalLength` or something similar which returns your suggestion from `PackageEntry`?

xPaw: That might be a good idea.

xPaw:
@babelshift Could you [squash](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes) so I can merge?

babelshift:
@xPaw Done, hopefully didn't screw anything up.

xPaw:
You managed to squash commits done in master, so your commit touches files that were already changed in master. You probably need to rebase.

babelshift:
That was my bad. I rebased on origin and not upstream, thus the merged commits were squashed.

xPaw:
Thanks for your work!

PR #32, opened 2015-11-30
Fix overflow when decompressing DXT1.

babelshift:
Fixes issue: https://github.com/SteamDatabase/ValveResourceFormat/issues/30


xPaw:
Thanks!

PR #35, opened 2015-12-16
Added initial version of KV3Reader

Perryvw:
Pretty sure this will fail for some cases, needs testing.


xPaw:
We might need to bring a kv1 implementation in first so that we can share common classes between all three versions. I was thinking on taking SteamKit's `KeyValues`, but due to license we'd need to ask all their contributors first.

SinZ163:
From what I can see, SteamKit's implementation wouldn't work to store KV3 (or KV2 even) data.

But this implementation **can** store kv1 (and almost KV2) but the process of turning it back to a string would need to be reworked to support their formattings

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,590 @@
...
+            currentString.Append(c);
+        }
+        
+        //Read a string value
+        private static void ReadValueString(char c)
+        {
+            if (c == '"')
+            {
+                //Multiline string detected
+                if (currentString.ToString() == "\"")
```
SinZ163: Wouldn't this treat the following as the start of a multistring and get confused
```
{
    emptyString = ""
    somethingElse = true
}
```

Note, Valves documentation states its only a start of a Multistring when [it starts with three unescaped quotes followed by an immediate newline (eg. the buffer looks like `\"\"\"\n`)](https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/KeyValues3#Multiline_String_Literals)

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,590 @@
...
+            }
+
+            currentString.Append(c);
+        }
+
+        //Reading multiline string
+        private static void ReadValueStringMulti(char c)
+        {
+            //Check for ending
+            if (c == '"' && currentString.ToString().Substring(currentString.Length - 2) == "\"\"")
```
SinZ163: Just a nitpick but valve docs state it also needs to have a `\n` there too.

[End string with a newline followed by three unescaped quotes (eg. the buffer looks like `\n\"\"\"`)](https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/KeyValues3#Multiline_String_Literals)

xPaw:
@SinZ163 whichever way it goes, I'd like to have all versions unified as much as possible.

SinZ163:
To go with the unification, it might be best to also move it into its own namespace (and folder), as its not directly part of the Resource family, which would allow for KV1 and KV2?? expansion

xPaw:
Sure.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+                if (str.Contains("\n"))
+                {
+                    stringBuilder.Append("\"\"\"");
+                    stringBuilder.Append(str);
+                    stringBuilder.Append("\"\"\"");
+                }
+                else
+                {
+                    stringBuilder.Append("\"");
+                    stringBuilder.Append(str);
```
SinZ163: Shouldn't this escape certain characters, so if my string value was `It's "alright"` it won't break

xPaw:
cc @netshroud for review

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+ * 
+ */
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.IO;
+using System.Globalization;
+
+namespace ValveKV
```
xPaw: This probably belongs under the `ValveResourceFormat.KeyValues` namespace. using KeyValues using KV in classname is a good idea too.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            {
+                stateStack.Pop();
+            }
+
+            currentString.Append(c);
+        }
+    }
+
+    //Data structure to contain the file contents
+    public class KVFile
```
xPaw: For future proofing, split each class into its own file (and put it in correct location, not under resourcetypes).

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+        public KVValue(Type type, Object value)
+        {
+            Type = type;
+            Value = value;
+        }
+    }
+
+    //Empty class to keep track of flagged strings
+    //This could probably be done better but meh.
+    public class FlaggedString { }
```
xPaw: This does feel like an unnecessary hack.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.IO;
+using System.Globalization;
+
+namespace ValveKV
+{
+    public static class KV3Reader
```
xPaw: Since you implement serialization too, it's not just a reader.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            //String opening
+            else if (c == '"')
+            {
+                stateStack.Pop();
+                stateStack.Push(State.VALUE_STRING);
+                currentString.Clear();
+            }
+            //Booleans
+            else if ((c == 'f' && (fileStream[index + 1] == 'a' && fileStream[index + 2] == 'l' && fileStream[index + 3] == 's' && fileStream[index + 4] == 'e'))
+                || (c == 't' && (fileStream[index + 1] == 'r' && fileStream[index + 2] == 'u' && fileStream[index + 3] == 'e')))
```
xPaw: How does Valve's parser react to case sensitivity here?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            stateStack = new Stack<State>();
+            stateStack.Push(State.HEADER);
+
+            root = new KVObject("root");
+            objStack.Push(root);
+
+            currentString = new StringBuilder();
+
+            //Open stream
+            fileStream = File.ReadAllText(filename).ToCharArray();
```
xPaw: Make a separate method that accepts a generic Stream, and make `Read(string)` open a stream for it. Reading the entire file and converting it to a char array is not a brilliant idea, use streams.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: Not if you assume the input is a KV file too since it should already escaped, you are right in the fact that if you put in a custom string this should probably be done, so depends on the usecase.

xPaw: It should escape the string correctly regardless of the use case.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: It is, as I indicated already, feel free to improve/suggest improvements.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: It was a stream first, that struggles with looking forward/back in C# though. Operating on a char array should realistically not make a difference.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
xPaw: Perhaps `KVValue` should keep track of its own `State` value?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
xPaw: Loading the entire file into memory is not nice. You can seek in streams if necessary.

Perryvw: Good point, will investigate.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
+/*
+ * KV3Reader class.
+ * This class reads in valve KV3 files and stores them in its datastructure.
+ *  
+ * Interface:
+ *  KVFile file = KV3Reader.ParseKVFile( fileName );
+ *  String fileString = file.Serialize();
+ *  
+ * TODO:
```
xPaw: Nit: it's unnecessary to add TODOs in comments like this, either mention them in the pull request, or create separate issues after its merged.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
+/*
+ * KV3Reader class.
+ * This class reads in valve KV3 files and stores them in its datastructure.
```
Netshroud: Nit: Valve have a capital V.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Netshroud: Agreed. This is not thread safe and no way to make it thread safe, unless you resign to never doing concurrent KV3 serialization.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+                stateStack.Push(State.SEEK_VALUE);
+                return;
+            }
+        }
+
+        //Seeking value state
+        private static void SeekValue(char c)
+        {
+            //Ignore whitespace
+            if (whiteSpace.Contains(c) || c == '=' )
```
Netshroud: `char.IsWhiteSpace` already exists. Is there some reason that you need to reimplement this?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Netshroud: Surely there's a better way to do this than char-by-char comparisons to the known literal?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+
+        //Read a boolean variable
+        private static void ReadValueBool(char c)
+        {
+            //Stop reading once the end of true or false is reached
+            if (c == 'e')
+            {
+                currentString.Append(c);
+                stateStack.Pop();
+                objStack.Peek().AddProperty(currentName, new KVValue(typeof(Boolean), currentString.ToString() == "true" ? true : false));
```
Netshroud: Break out the actual value into a separate variable.
Nit: use `typeof(bool)` instead of `typeof(Boolean)`.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            }
+
+            currentString.Append(c);
+        }
+
+        //Read a boolean variable
+        private static void ReadValueBool(char c)
+        {
+            //Stop reading once the end of true or false is reached
+            if (c == 'e')
```
Netshroud: This feels hacky.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+        {
+            //For arrays
+            if (c == ',')
+            {
+                stateStack.Pop();
+                stateStack.Push(State.SEEK_VALUE);
+                if ( currentString.ToString().Contains('.')) {
+                    objStack.Peek().AddProperty(currentName, new KVValue(typeof(Double), Double.Parse(currentString.ToString(), CultureInfo.InvariantCulture)));
+                } else {
+                    objStack.Peek().AddProperty(currentName, new KVValue(typeof(Int32), Int32.Parse(currentString.ToString(), CultureInfo.InvariantCulture)));
```
Netshroud: Prefer `typeof(double)` and `typeof(int)`, as above with `bool`.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+                return;
+            }
+
+            currentString.Append(c);
+        }
+
+        //Read an array
+        private static void ReadValueArray(char c)
+        {
+            //This shouldn't happen
```
Netshroud: If this shouldn't happen, why is this a `Console.WriteLine` instead of an exception? And why use `Console.WriteLine`?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            {
+                currentString.Append(c);
+            }
+        }
+
+        //Read a comment block
+        private static void ReadCommentBlock(char c)
+        {
+            //Look for the end of the comment block
+            if (c == '/' && currentString.ToString().Last() == '*')
```
Netshroud: I'm seeing `currentString.ToString()` everywhere. Evaluate the performance impact - there's probably a faster way than assembling a string, scanning the string, and throwing it away every time.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+        public KVFile(String header, KVObject root)
+        {
+            this.Header = header;
+            this.Root = root;
+        }
+
+        //Serialize the contents of the data structure
+        public String Serialize()
+        {
+            StringBuilder stringBuilder = new StringBuilder();
```
Netshroud: Nit: `var`

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            }
+
+            currentString.Append(c);
+        }
+    }
+
+    //Data structure to contain the file contents
+    public class KVFile
+    {
+        public String Header;
```
Netshroud: Why are these public? Consider making this `readonly`, with a public getter if necessary.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            stringBuilder.AppendLine(Header);
+            Root.Serialize(stringBuilder, 1);
+            return stringBuilder.ToString();
+        }
+    }
+
+    //Datastructure for a KV Object
+    public class KVObject
+    {
+        public String key;
```
Netshroud: As above

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+        public KVObject Root;
+
+        public KVFile(String header, KVObject root)
+        {
+            this.Header = header;
+            this.Root = root;
+        }
+
+        //Serialize the contents of the data structure
+        public String Serialize()
```
Netshroud: Nit: Prefer `string` over `String`, as above with other types.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            {
+                stringBuilder.Append(((Double)value).ToString("#0.000000",CultureInfo.InvariantCulture));
+            }
+            else if (type == typeof(Int32))
+            {
+                stringBuilder.Append((Int32)value);
+            }
+            else
+            {
+                //This shouldn't happen
```
Netshroud: So do something about it, rather than passively swallowing known error cases.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+                stringBuilder.Append(value);
+            }
+        }
+    }
+
+    //KV Array extending the KVObject
+    public class KVArray : KVObject
+    {
+        new public List<KVValue> Properties;
```
Netshroud: What?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+
+        public KVArray(String name) : base(name)
+        {
+            Properties = new List<KVValue>();
+        }
+
+        public override void AddProperty(String name, KVValue value)
```
Netshroud: `name` is silently discarded here. Should `KVArray` really be a subclass of `KVObject` (which acts more like `KVDictionary`)?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+        {
+            Properties.Add(value);
+        }
+
+        public override void Serialize(StringBuilder stringBuilder, int indent)
+        {
+            stringBuilder.Append(Environment.NewLine);
+            for (var i = 0; i < indent-1; i++)
```
Netshroud: Second or third time I'm seeing this - break it out into a function.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            {
+                stringBuilder.Append("\t");
+            }
+            stringBuilder.Append("[");
+            stringBuilder.Append(Environment.NewLine);
+
+            foreach (KVValue entry in Properties) {
+                for (var i = 0; i < indent; i++)
```
Netshroud: Inconsistent bracing style

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            for (var i = 0; i < indent - 1; i++)
+            {
+                stringBuilder.Append("\t");
+            }
+            stringBuilder.AppendLine("]");
+        }
+    }
+
+    //Class to hold type + value
+    public class KVValue
```
Netshroud: Is `KVValue<T>` feasible here, instead of what amounts to a `Tuple<Type, object>`?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+    {
+        public Type Type;
+        public Object Value;
+
+        public KVValue(Type type, Object value)
```
Netshroud: Nit: Prefer `object` over `Object`

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            }
+            //Number
+            else if (numerals.Contains(c))
+            {
+                stateStack.Pop();
+                stateStack.Push(State.VALUE_NUMBER);
+                currentString.Clear();
+                currentString.Append(c);
+            }
+            //Flagged resource
```
Netshroud: Code doesn't explain what a flagged resource is, just that it's a `FlaggedString` or a `string` - I'm unclear on what this is, judging by this PR, and having little experience with KeyValues3. Either explain with a comment, or let the code do the talking.

Also, is it guaranteed for all eternity that anything that is not the above is a flagged resource, or is there some identifier you can use to detect a flagged resource specifically?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+            else if ((c == 'f' && (fileStream[index + 1] == 'a' && fileStream[index + 2] == 'l' && fileStream[index + 3] == 's' && fileStream[index + 4] == 'e'))
+                || (c == 't' && (fileStream[index + 1] == 'r' && fileStream[index + 2] == 'u' && fileStream[index + 3] == 'e')))
+            {
+                stateStack.Pop();
+                stateStack.Push(State.VALUE_BOOL);
+                currentString.Clear();
+                currentString.Append(c);
+            }
+            //Number
+            else if (numerals.Contains(c))
```
Netshroud: Consider `char.IsNumber` if you haven't already, but take Unicode numbers that aren't `[0-9]` into consideration.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,592 @@
...
+                objStack.Peek().AddProperty(currentName, new KVValue(typeof(Boolean), currentString.ToString() == "true" ? true : false));
+                return;
+            }
+
+            currentString.Append(c);
+        }
+
+        //Read a numerical value
+        private static void ReadValueNumber(char c)
+        {
```
Netshroud: Does this numeric implementation allow `345-678`, since '-' is in the set of numeric characters you've defined? I assume that would crash your parser? '-' should probably be only allowed if it's leading, as a unary negation operator.

Netshroud:
cc successful. :smile: 

I'm concerned that for something of this complexity there seem to be no unit tests.

Also, can I recommend something like [StyleCopAnalyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers) to enforce a standard style for this project? Even within this one file the styling is inconsistent.

loveolsson:
This isn't really related, because it's written in node.js, but anyway we had to parse KV3 files in a different project and i wrote this really quick and dirty parser that converts it to a JS-object. 
https://github.com/loveolsson/KV3toJS
I've used your project for reference to I thought I'd throw this in there if it's of any use for anyone.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: Changed namespace in next version, still unsure what the actual class name should be.

xPaw: Just do KeyValues3 for now.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: Switched to a KVType enum instead of doing typeof stuff each time for the next commit.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: Removed this in clean-up, everything is a KVObject now that just tracks if it is an array or not.

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
Perryvw: Haven't written C# in a few years, wrote a quick parser on SinZ's request. I was not even aware that function existed. Changed to char.IsWhiteSpace.

SinZ163:
with FlaggedString, it might be better for it to be its own class again, and having a Flag field and Value field, so `stringThatIsAResourceReference = resource:"particles/items3_fx/star_emblem.vpcf"` would end up as
``` csharp
flagValue.Flag == "resource";
flagValue.Value == "particles/items3_fx/star_emblem.vpcf";
```
rather than its current implementation being
``` csharp
flagValue == "resource:\"particles/items3_fx/star_emblem.vpcf\""
```

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
SinZ163: From memory it was originally a generic at one point, but became annoying to work with as you cant have something like a `Dictionary<string, KVType>` anymore as `KVType<bool>` and `KVType<string>` are considered different types, and if you have a superclass for it so it does fit,the usefulness of generics start going away if you have to constantly type cast INTO the generic

 But this may just be me and Perry actually being horrible at C#

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,627 @@
...
+
+            //Need to preserve the order
+            for (var i = 0; i < Count; i++)
+            {
+                for (var j = 0; j < indent; j++)
+                {
+                    stringBuilder.Append("\t");
+                }
+
+                PrintValue(stringBuilder, Properties[i.ToString()].Type, Properties[i.ToString()].Value, indent);
```
SinZ163: Why are you splitting KVValue up into its two fields rather than just sending the one in?

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,627 @@
...
+                case KVType.FLAGGED_STRING:
+                    stringBuilder.Append((string)value);
+                    break;
+                case KVType.STRING:
+                    stringBuilder.Append("\"");
+                    stringBuilder.Append((string)value);
+                    stringBuilder.Append("\"");
+                    break;
+                case KVType.STRING_MULTI:
+                    stringBuilder.Append("\"\"\"");
```
SinZ163: Needs the \n to be Valve Documentation accurate STRING_MULTI

ValveResourceFormat/Resource/ResourceTypes/KV3.cs
```diff
@@ -0,0 +1,627 @@
...
+                    stringBuilder.Append((string)value);
+                    stringBuilder.Append("\"\"\"");
```
SinZ163: ^

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,478 @@
...
+            {
+                parser.stateStack.Pop();
+            }
+
+            parser.currentString.Append(c);
+        }
+    }
+
+    //Data structure to contain the file contents
+    public class KVFile
```
xPaw: Should've moved it into separate file too.

ValveResourceFormat/KeyValues/KVObject.cs
```diff
@@ -0,0 +1,193 @@
...
+                    break;
+                default:
+                    //Unknown type encountered
+                    throw new Exception("Unknown type encountered.");
+            }
+        }
+    }
+
+    //Class to hold type + value
+    public class KVValue
```
xPaw: We can afford having each enum in a separate file, just do that.

Netshroud:
Why all the public fields?

Also, I'm not familiar with whatever style ValveResourceFormat is using - I've only briefly scanned over the project - but the keyword `private` is almost never needed.

ValveResourceFormat/KeyValues/KVObject.cs
```diff
@@ -0,0 +1,217 @@
...
+        //Serialize the contents of the KV object
+        private void SerializeObject(StringBuilder stringBuilder, int indent)
+        {
+            //Don't enter the top-most object
+            if (indent > 1)
+            {
+                stringBuilder.Append("\n");
+            }
+
+            for (var i = 0; i < indent - 1; i++)
```
Netshroud: This indentation-writing loop is everywhere. Consider refactoring it out into a function or extension method.

ValveResourceFormat/KeyValues/KVObject.cs
```diff
@@ -0,0 +1,217 @@
...
+                    break;
+                case KVType.DOUBLE:
+                    stringBuilder.Append(((double)value).ToString("#0.000000", CultureInfo.InvariantCulture));
+                    break;
+                case KVType.INTEGER:
+                    stringBuilder.Append((int)value);
+                    break;
+                default:
+                    //Unknown type encountered
+                    throw new Exception("Unknown type encountered.");
```
Netshroud: Don't throw `System.Exception`. Either pick an appropriate exception type (`InvalidDataException`/`ArgumentException`/`ArgumentOutOfRangeException`/`InvalidOperationException`) or write your own.

ValveResourceFormat/KeyValues/KVObject.cs
Netshroud: For consideration: Should `KVValue` be mutable, or immutable? If it should be immutable, make `Type` and `Value` read-only properties.

Also for consideration: As such a simple type, can `KVValue` be a struct, or must it be an object?

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+                return ParseKVFile(fileStream);
+            }
+        }
+
+        public static KVFile ParseKVFile(Stream fileStream)
+        {
+            var parser = new Parser();
+
+            //Open stream reader
+            parser.fileStream = new StreamReader(fileStream);
```
Netshroud: This seems like `fileStream` should be an immutable field set by the constructor. If it really really really should be a mutable property, make it PascalCase, not camelCase.

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+                parser.stateStack.Pop();
+                
+                //Can directly be added
+                parser.objStack.Peek().AddProperty(parser.currentName, new KVValue(KVType.BOOLEAN, false));
+
+                //Skip next characters
+                SkipChars(parser, 4);
+            }
+            //Boolean true
+            else if (c == 'f' && PeekString(parser, 3) == "rue")
```
Netshroud: `frue`? Looks like a copy/paste bug. Perhaps write a function that combines the current char with your read-ahead, so that you can just say `if (ReadAheadMatches("true")) { ... }`

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+            {
+                parser.stateStack.Pop();
+
+                //Can directly be added
+                parser.objStack.Peek().AddProperty(parser.currentName, new KVValue(KVType.BOOLEAN, true));
+
+                //Skip next characters
+                SkipChars(parser, 3);
```
Netshroud: Magic number. Replace with something like `"true".Length - 1` or roll it into the function I suggested above.

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+            parser.currentString.Append(c);
+        }
+
+        //Read an array
+        private static void ReadValueArray(char c, Parser parser)
+        {
+            //This shouldn't happen
+            if (!Char.IsWhiteSpace(c) && c != ',')
+            {
+                throw new Exception("Error parsing array.");
```
Netshroud: Same issue with exceptions - pick or write an exception type.

ValveResourceFormat/KeyValues/KeyValues3.cs
Netshroud: Same questions on mutability as `KVObject`.

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+        public KVObject Root;
+
+        public KVFile(String header, KVObject root)
+        {
+            this.Header = header;
+            this.Root = root;
+        }
+
+        //Serialize the contents of the data structure
+        public string Serialize()
```
Netshroud: Does this need to be it's own function, or would it be more appropriate to override `ToString`?

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+            for (int i = 0; i < num; i++)
+            {
+                NextChar(parser);
+            }
+        }
+
+        //Utility function
+        private static string PeekString(Parser parser, int length)
+        {
+            char[] buffer = new char[length];
```
Netshroud: Why the buffering? Are you explicitly attempting to support `Stream`s that don't support seek operations, or is this just overengineering?

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,526 @@
...
+                {
+                    buffer[i] = parser.charBuffer.ElementAt(i);
+                }
+                else
+                {
+                    buffer[i] = (char)parser.fileStream.Read();
+                    parser.charBuffer.Enqueue(buffer[i]);
+                }
+            }
+            return String.Join("", buffer);
```
Netshroud: Prefer `string` over `String`, and `string.Empty` over `""`.

ValveResourceFormat/KeyValues/KeyValues3.cs
Perryvw: Yes, the idea is to avoid seeking (which might be overengineering, but I don't know what the input exactly is myself).

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,535 @@
...
+            VALUE_ARRAY,
+            VALUE_STRING,
+            VALUE_STRING_MULTI,
+            VALUE_NUMBER,
+            VALUE_FLAGGED,
+            COMMENT,
+            COMMENT_BLOCK
+        }
+
+        private class Parser
```
xPaw: I still don't see the need in having a separate parser class. Just make `KeyValues3` non static, and you can still leave all reading methods static within the same class.

ValveResourceFormat/KeyValues/KeyValues3.cs
```diff
@@ -0,0 +1,535 @@
...
+            //Start looking for the next property name
+            parser.stateStack.Push(State.PROP_NAME);
+            parser.currentString.Clear();
+            parser.currentString.Append(c);
+        }
+
+        //Read a string value
+        private static void ReadValueString(char c, Parser parser)
+        {
+            if (c == '"' && parser.previousChar != '\\')
```
xPaw: How does Valve's parser handle escape sequences for chars other than `"`? What would happen if it's `\0` or `\40`?

xPaw:
I'm not a fan of how strings are read. Ideally `charBuffer` and `previousChar` would go away.

Issue #29, opened 2015-11-28
vtex_c files which contain TGA compressed with DXT5 have incorrect colors when decompressed


babelshift:
After some testing and research, I've found that the vtex_c files which are incorrectly colored on extraction are TGA files compressed with DXT5 and possibly in YCoCg colorspace.

Looking at the REDI tab reveals:

CResourceString m_String = "Texture Compiler Version Image YCoCg Conversion"

This is a bit out of my league. I'm not sure if Valve's texture compiler is converting to YCoCg colorspace prior to compression. Maybe that means it needs to be read out as YCoCg and converted to RGB during decompression?

Github doesn't support attaching vtex_c, but here's a test file that you can see. https://drive.google.com/file/d/0B7o2acNuwX44X01LZnVUeFZfX2c/view?usp=sharing

Attached to this issue is the results of decompressing the file linked above.
![loadingscreen_tga](https://cloud.githubusercontent.com/assets/2723423/11454121/7978eb00-95f0-11e5-831e-e69ad2013569.png)

SinZ163:
Part of the problem is the use of alpha, and ValveResourceViewer has a white background (dota always has a black background).  But there is still a colour issue.

babelshift:
The alpha channel seems off in addition to the colors. I saved it out of ValveResourceViewer as a png and viewed it in Paint.NET. Colors and alpha looked weird there as well. I played around with converting between colorspaces with ImageMagick, but I don't think it supports YCoCg.

xPaw:
There has to be some information besides compiler identifier stored for this. Have you tried finding a DXT5 file with RGB color scheme, if one exists?

babelshift:
I did look but only on a limited (manual) basis. I might try to create something that goes through the vpk and checks for the compression and color spaces automatically.

babelshift:
Some info from this thread which might be relevant: http://dev.dota2.com/showthread.php?t=182677&page=2

> Okey.
 
> Good news. The format used`s definitely YCgCo (https://en.wikipedia.org/wiki/YCgCo). I`ve managed to compile a YCgCo formatted image using it`s Valve-spinoff, and my Dota 2 client displayed it correctly. The schema Valve`s using`s the following:
> R: Co
> G: Cg
> B: Edit: Nope, it`s not Alpha. It`s .. something. Can`t discern what, tho.
> A: Y

xPaw:
[...]

PR #36, opened 2015-12-17
Modified to handle DXT5 color correctly, create straight alpha and add alpha to bitmap output

loveolsson:


coral:
Great, had the issue with some textures. This solves it for me!

xPaw:
Nice! Would you mind showing a couple of example files which were incorrectly rendered before?

loveolsson:
This is /material/particle/lens_flare/lens_flare.vtex_c, I've but it on a gray background to better show of the alpha.

Before:
lens_flare_old

With unmodified alpha:
lens_flare_premultiplied

With straight alpha:
lens_flare_straight

loveolsson:
According to https://en.wikipedia.org/wiki/S3_Texture_Compression#DXT4_and_DXT5 DTX5 is already straight alpha, but the output was clearly premultiplied, so I don't know what's going on there. I went for good looking output from Dota 2 rather than believing what I read.

xPaw:
That certainly looks better, but going for "good looking ouput" might be not the best idea. Have you looked at ffmpeg's implementation to perhaps see how they handle it?

coral:
I was able to use it with this patch as is when toying around with particles so it doesn't only look right, it behaves right. Does FFMPEG handle DXT5? That is a great suggestion if it does to look at how they do it. @xPaw

xPaw:
ffmpeg does support DXT5 as I had to look at their implementation to handle YCoCg colorscheme.

loveolsson:
Well, yes. I removed the YCoCg handling from the code, because it just didn't behave right for the Dota textures, instead it behaved exactly like DTX1 colors. Could it be that YCoCg is only used for video implementations?

xPaw:
See #29

coral:
Does this mean they are using different colorspaces in different textures?

xPaw:
They could be, but there is no direct way of checking it. Perhaps it's related to alpha itself?

loveolsson:
Iinteresting. So basically this should be used if 
ResourceSpecialDependency_t
CResourceString m_String = "Texture Compiler Version Image YCoCg Conversion"
is in the REDI?

xPaw:
Probably, but looking at REDI to determine the color scheme is going to bite us in the future for sure. We need to know how Valve handles it on their side.

coral:
@xPaw i will dig a bit more into this.

ValveResourceFormat/ThirdParty/DDSImage.cs
```diff
@@ -229,47 +229,76 @@ private static void DecompressBlockDXT5(int x, int y, int width, byte[] blockSto
                         }
                     }
 
-                    byte colorCode = (byte)((code >> 2 * (4 * j + i)) & 0x03);
-
+                    byte positionCode = (byte)((code >> 2 * (4 * j + i)) & 0x03);
                     int finalR = 0, finalG = 0, finalB = 0;
 
-                    switch (colorCode)
+                    if (color0 > color1)
```
SinZ163: Rather than having an if statement with two switch cases, couldn't the if statement just increment positionCode by 4, and have:
```csharp
switch(positionCode)
{
    case 0:
    case 4:
        //INSERT LOGIC
        break;
    case 1:
    case 5:
        //insert logic
        break
    case 2:
        break
    case 3:
        break
    case 6:
        break
    case 7:
        break
}
```

loveolsson:
So, for reference this is what the YCgCo implementation is based on:
http://www.nvidia.com/object/real-time-ycocg-dxt-compression.html

The bad news is that we've done a bit more research on how to differentiate where YCgCo is used, and everything point to that is just down to the dependency in the metadata if the conversion is applied.
I've moved straight alpha and YCgCo conversion to functions in Texture.cs that performs the conversion on the existing bitmap and returns a new bitmaps. I's suggest these being implemented as command line options on the Decompiler.

ValveResourceFormat/ThirdParty/DDSImage.cs
loveolsson: The color portions of DXT5 and DXT1 is identical and could also be unified. I just copied the DXT1 implementation because is was working correctly.

loveolsson:
By the way, FFMPEG reads the FOURCC from the video container to determine what format to use. 
https://github.com/FFmpeg/FFmpeg/blob/603ebab8d7b789cd3cb6c7ca645a7badb5b950b4/libavcodec/dds.c#L128

SinZ163:
after button mashing through other images, theres no vtex flag / extra_data for it, as lens flare in question has none, It might have initially been mipinfo stuff, but I found images without mipinfo but with this colour problem. So the only correlation that hasn't been falsified has been the REDI tag.

It would be preferred to have the colour correction stuff handled in the DDSImage class purely so you aren't double looping over every pixel ever, and in some high res images potentially millions of times, maybe feeding in some flag int or something to the DXT5 functions?

loveolsson:
I agree.

Question is how far to feed it, because it has to travel thru for example GenerateBitmap where a DXTOptions input makes little sense. Any suggestions?

xPaw:
I see no problem feeding parameters into dxt5 method.

loveolsson:
Yeah, sorry. Meant to do a proper new fork out of this. My mistake. Basically, we need a version for right now that does what we need it to for the project we work on. This was never meant as a proper push, I messed up my git push.

SinZ163:
then rebase the pullrequest branch to when it was last acceptable?

loveolsson:
I don't know if it ever was. It basically messed up a lot of your previous work and I don't want to accidentally pollute your code with my quick fixes.

xPaw:
You don't pollute our code until we accept the pull request. You can reset your branch to our master and re-implement the fix. We still need to figure out when to apply YCoCg color conversion for this to work correctly.

SinZ163:
The latest commit was *almost* the proposed solution.

Just rather than having the logic to choose what colourspace in `Decompiler/Program.cs` have it handled inside Texture.GenerateBitmap() by iterating over [the REDI Special Dependancies](https://github.com/SteamDatabase/ValveResourceFormat/blob/master/ValveResourceFormat/Resource/Resource.cs#L134) until you see `Texture Compiler Version Image YCoCg Conversion` as the `String` value. And if you don't, then assume its in the RGB Colour space as you weren't told otherwise.

And then just feed that enum value which you made in your last commit OR a boolean into `UncompressDXT5`

xPaw:
@SinZ163 is right, a boolean will do enough. And there shouldn't be a need for `StraightAlpha` option.

loveolsson:
Messed that up bad, haha. Don't ask. Seriously, don't. I am bad and I feel bad. End result is that I made a new fork with the scaled back adjustments. 
https://github.com/loveolsson/ValveResourceFormat

loveolsson:
Wouldn't SUGGEST_CLAMPU most likely be a texture clamp mode reference?

PR #41, opened 2015-12-19
DXT5 reads RGB but can do YCgCo conversion and create straight alpha

loveolsson:
New pull request that replaces
https://github.com/SteamDatabase/ValveResourceFormat/pull/36

UncompressDXT5 takes two extra bools that enable YCgCo conversion and straight alpha.

### DIFF
--- a/ValveResourceFormat/ThirdParty/DDSImage.cs
+++ b/ValveResourceFormat/ThirdParty/DDSImage.cs
@@ -270,7 +270,9 @@ namespace ValveResourceFormat.ThirdParty
                             finalAlpha = byte.MaxValue;
                         } else if (finalAlpha > 0)
                         {
-
+                            finalR = ClampColor(finalR * byte.MaxValue / finalAlpha);
+                            finalG = ClampColor(finalG * byte.MaxValue / finalAlpha);
+                            finalB = ClampColor(finalB * byte.MaxValue / finalAlpha);
                         }
 
                         image.SetPixel(x + i, y + j, Color.FromArgb(finalAlpha, finalR, finalG, finalB));
--- a/ValveResourceFormat/ThirdParty/DDSImage.cs
+++ b/ValveResourceFormat/ThirdParty/DDSImage.cs
@@ -266,9 +266,14 @@ namespace ValveResourceFormat.ThirdParty
                             finalR = ClampColor(finalAlpha + co - cg);
                             finalG = ClampColor(finalAlpha + cg);
                             finalB = ClampColor(finalAlpha - co - cg);
+
+                            finalAlpha = byte.MaxValue;
+                        } else if (finalAlpha > 0)
+                        {
+
                         }
 
-                        image.SetPixel(x + i, y + j, Color.FromArgb(finalR, finalG, finalB));
+                        image.SetPixel(x + i, y + j, Color.FromArgb(finalAlpha, finalR, finalG, finalB));
                     }
                 }
             }


xPaw:
You didn't rebase on master correctly, thus your diffs include unnecessary changes.

What's the need in straight alpha option?

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -161,11 +163,11 @@ public Bitmap GenerateBitmap()
                             }
                         }
 
-                        return ThirdParty.DDSImage.UncompressDXT5(Reader, Width, Height);
+                        return ThirdParty.DDSImage.UncompressDXT5(Reader, Width, Height, ycgco, Flags.HasFlag(VTexFlags.SUGGEST_CLAMPU));
```
xPaw: You're not passing the options in correctly. `ycgco` is never true

loveolsson:
I could be reading everything completely wrong here, but how do I gain access to the Flags in the instance of Texture from the static DDSImage class?

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
loveolsson: I know, it's prepared for the discussed function in GenerateBitmap that would read the REDI data, but just now it's just set to false.

xPaw:
@SinZ163 is saying he found a YCoCg image that had no vtex flags, so CLAMPU fails for this purpose. Looks like we will need to look at REDI strings after all.

As for the straight alpha option, could you explain the need in doing that?

xPaw:
Basically to look at REDI you need to do something similar to this: https://github.com/SteamDatabase/ValveResourceFormat/blob/master/ValveResourceFormat/Resource/Resource.cs#L132-L140
but instead of looking at first item in the list, iterate all of them, and check if one of them matches the exact YCoCg compiler string.

loveolsson:
Oh, I missread https://github.com/SteamDatabase/ValveResourceFormat/pull/36#issuecomment-166028138 I thought you meant that controlled straight alpha. 

Anyway. The reason the alpha option is needed in some way or another is basically that PNG only support straight alpha (where a pixel is proportionally blended with the background based on alpha value) and not premultiplied alpha (where the alpha is cut from the background each pixel added on top).

If you use a PNG with wrong alpha you get a dark halo around everything. But the crappy part is that the process of making the alpha straight is lossy. I guess that is why Bitmap.Save method don't do it even for premultiplied pixel formats.
The DXT5 specification says the format is straight alpha, but all the textures I've looked at from Dota 2 looks to be premultiplied (DXT4).

So, basically the choice is between making the PNGs usable or make them as lossless/conforming as possible. I don't think a command line flag for that is such a bad idea.

loveolsson:
I haven't really had the time to properly look into REDI-reading yet, but from the first quick looks I don't understand how to access other blocks from inside the Texture class. Next week I'll have more time to sit down and understand it all, but a push in the right direction would be good.

I understand how to do it anywhere I have access to the resource-object where I have Blocks to loop thru, but not inside Texture.

xPaw:
I did that for you in 1220a94b20a70ca4ad3d4ce925d383003f3eae78

xPaw:
Just tried to mess with this, looks like some textures that have blue background are turned into alpha in png. That's how it's supposed to work, right?

@SinZ163 what do you think?

PR #80, opened 2016-11-28
Update to latest

ihatecsv:


ihatecsv:
I forgot how Github worked for a second there, pls ignore

PR #81, opened 2016-12-13
Adding export_filter and from_vpk_directory as command line arguments

DarkMio:
This PR adds two new options to the command line decompiler:

``--from_vpk_directory some\path`` matches all directories that start with the given filter (and the impl is lazy)  
``--export_filter filter`` only matches files in which the filename is contains the given string (simliarly lazy)

This lets you target specific files and folders for that matter. There is no way supported to have multiple folders for now.

### DIFF
--- a/Decompiler/Program.cs
+++ b/Decompiler/Program.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


Decompiler/Program.cs
```diff
@@ -535,6 +539,7 @@ private static void ParseVPK(string path)
             else
             {
                 Console.WriteLine("--- Dumping decompiled files...");
+
```
SinZ163: Probably want to add another option to force a DumpVPK of a specified file extension

xPaw:
Superseded by #87.

PR #86, opened 2017-07-22
Add -f [filter] -e [extension] -s [silent] dump options

AveYo:
HLExtract supports such filtering, but that means extra intermediary files being generated
Decompiler is not very useful without it when it comes to processing DOTA pak01_dir.vpk

### DIFF
--- a/Decompiler/Program.cs
+++ b/Decompiler/Program.cs
@@ -699,7 +699,24 @@ namespace Decompiler
                                     }
                                     break;
                                 default:
-                                    output = Encoding.UTF8.GetBytes(resource.Blocks[BlockType.DATA].ToString());
+                                    try
+                                    {
+                                        output = Encoding.UTF8.GetBytes(resource.Blocks[BlockType.DATA].ToString());
+                                    }
+                                    catch (Exception e)
+                                    {
+                                        if (!Options.Silent)
+                                        {
+                                            lock (ConsoleWriterLock)
+                                            {
+                                                Console.ForegroundColor = ConsoleColor.DarkMagenta;
+                                                Console.WriteLine("--- Resource type " + type + " decompiler not implemented, extracting as-is..." );
+                                                Console.ResetColor();
+                                            }
+                                        }
+                                        package.ReadEntry(file, out output);
+                                        newType = type;
+                                    }
                                     break;
                             }
                         }                       


Decompiler/Program.cs
```diff
@@ -115,6 +129,11 @@ private static void ProcessFile(string path)
             {
                 ParseVPK(path);
 
+                if (File.Exists(path + ".manifest.txt"))
+                {
+                    File.Delete(path + ".manifest.txt");
```
xPaw: Why are you deleting the manifest here? Deleting it like this renders it useless.

xPaw:
Basically this seems to be conflicting with the hardcoded dumping we have. If you could remove that hardcoded extensions array and make sure it can be done with parameters alone, that would be much better.

Decompiler/Program.cs
AveYo: The manifest.txt is a great addition, I only removed it temporary to force a dump and forgot about it - blaming it on coding Valve-style...  in a text editor :)

AveYo:
Decompiler is quite fast with filtering! 
Take for example decompiling all the particles from DOTA `pak01_dir.vpk` _(~1k folders, ~30k files)_:
Using **HLExtract** for extracting with filtering, and then feeding the output folder as input to **Decompiler**, it runs around `6+` minutes _(removing tmp files_).
By adding filtering to **Decompiler**, it runs the same task in less than `2` minutes and does not trash the hdd with temporary files.
`decompiler -i "%dotavpk%" -f particles\ -o "%dotacontent%" -s`
00:02:05,24
`decompiler -i "%dotavpk%" -f particles\ -e vpcf_c -o "%dotacontent%" -s`
00:01:37,28 
Filtering also by extension gains significant speed by not processing other file types.

**Thank you for creating such an amazing library and tools!**  

_I would also like to formally ask for permission to include the Decompiler binaries in my DOTA No-Bling project_

xPaw:
Currently the dump vpk option is basically done for GameTracker, where it only extracts and decompiles hardcoded file types (which you tried moving). Ideally we would not have it hardcoded, and simply provide what we want to do via CLI (and we probably want an extra param saying if it should be decompiled or not).

Basically I'm fine with overhauling VPK CLI as long as we can make it work as the current solution in master (so that GameTracker still works).

AveYo:
Sorry, I have nuked the original repository - I'm not very familiar with github.
Please close this "pull request", as I've created [another one](https://github.com/SteamDatabase/ValveResourceFormat/pull/87)  
So I've refined it a little, Decompiler can now be used as a complete vpk command line tool without needing HLExtract for doing partial extraction from game files. 
It should also not conflict with default **GameTracker** usage, as the extra resource types are only activated by adding **--tool** option (filtering works regardless). Also changed **-e** to accept multiple extensions.
Now back to my No-Bling DOTA particles mod 👍

xPaw:
There is never a reason to create a new PR if you fuck up, the current one can always be fixed, FYI.

Decompiler/Program.cs
```diff
@@ -449,9 +449,12 @@ private static void ParseVPK(string path)
...
             {
-                Console.ForegroundColor = ConsoleColor.Green;
-                Console.WriteLine("--- Listing files in package \"{0}\" ---", path);
-                Console.ResetColor();
+                if (!Options.Silent)
+                { 
+                		Console.ForegroundColor = ConsoleColor.Green;
+                		Console.WriteLine("--- Listing files in package \"{0}\" ---", path);
+                		Console.ResetColor();
+            		}
```
xPaw: Fix whitespace here.

Decompiler/Options.cs
```diff
@@ -42,6 +42,26 @@ class Options
...
+
+        [Option('f', "filter",
+            HelpText = "Dump filter starting with path, ex: panorama\\")]
+        public string Filter { get; set; }
+
+        [Option('e', "filter_ext",
+            HelpText = "Dump filter extension(s), example: vcss_c,vjs_c,vxml_c")]
+        public string FilterExt { get; set; }       
+
+        [Option('s', "silent",
```
xPaw: Can we leave this silent option for another PR, will make it easier to review this.

Decompiler/Options.cs
```diff
@@ -42,6 +42,26 @@ class Options
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
+        [Option("tool", DefaultValue = false,
+            HelpText = "Enable generic vpk dump tool for all supported resources.")]
+        public bool VPKTool { get; set; }
```
xPaw: What's the point of this option? If you wanted to keep gametracking compat, I suggested that removing that hardcoded code is fine as long as the same behaviour can be achieved using CLI only (extract + filter + dump).

Decompiler/Options.cs
```diff
@@ -42,6 +42,26 @@ class Options
...
+
+        [Option('x', "extract",
```
xPaw: Can you prefix all these extract/filter options with `vpk_` please.

Decompiler/Program.cs
```diff
@@ -596,19 +610,25 @@ private static void ParseVPK(string path)
...
             if (!package.Entries.ContainsKey(type))
             {
-                Console.WriteLine("There are no files of type \"{0}\".", type);
+                if (!Options.Silent) Console.WriteLine("There are no files of type \"{0}\".", type);
 
                 return;
             }
 
             var entries = package.Entries[type];
+            uint oldCrc32;
```
xPaw: Why did you add this back?

Decompiler/Program.cs
```diff
@@ -621,41 +641,85 @@ private static void DumpVPK(Package package, string type, string newType)
...
 
-                if (type.EndsWith("_c", StringComparison.Ordinal))
+                if (type.EndsWith("_c", StringComparison.Ordinal) && !Options.Extract)
                 {
                     using (var resource = new Resource())
                     {
                         using (var memory = new MemoryStream(output))
                         {
                             resource.Read(memory);
-                        }
```
xPaw: Why did you move the switch inside of this using statement? This doesn't appear to be necessary.

Decompiler/Options.cs
AveYo: Well, first, I had no idea where this tool is used ([here?](https://github.com/SteamDatabase/GameTracking/blob/master/common.sh)), and second, I followed the suggestion to have independent cli option.   
But if it's only used there, then I guess it makes sense to drop the hard-coded options, and add in common.sh: `-e "vxml_c,vjs_c,vcss_c,vsndevts_c,vpcf_c,txt,cfg,res,png,jpg"`

Decompiler/Program.cs
AveYo: Won't compile on Visual Studio Community Edition 2015, will remove it next edit.

Decompiler/Options.cs
AveYo: Will do.

Decompiler/Options.cs
AveYo: I only added it for speed since even redirecting to null slows the process when there are 30k+ files. Will do.

Decompiler/Program.cs
AveYo: It is necessary for extracting sound resources, as the memory memorystream would be closed and GetSound() would fail. I guess you would prefer creating a new stream but that means more code. 
There is no harm in keeping the stream open until work with the resource is done?!

Decompiler/Options.cs
xPaw: That is precisely what I was suggesting!

Decompiler/Program.cs
xPaw: It's a C#7 feature

Decompiler/Program.cs
xPaw: Ah okay makes sense.

Decompiler/Program.cs
AveYo: Looks like AppVeyor still uses an older version and fails the build, too.
What's the harm in keeping it VS2015-compatible?
```c#
uint oldCrc32;
if (OldPakManifest.TryGetValue(filePath, out oldCrc32) && oldCrc32 == file.CRC32)
```
Well, I can see one -  restraining from using this coding style in the future :)

PR #87, opened 2017-07-23
Generic vpk dump tool functionality and filters

AveYo:
Revised filter options:
-d , --vpk_decompile : Decompile supported files (bool)
-e , --vpk_extensions : File extension(s) filter, example: vxml_c,vjs_c or "vxml_c,vjs_c"
-f , --vpk_filepath : File path filter, example: panorama/ or panorama\ or "panorama\\\\"

Generic vpk decompiler will fallback to extract-only for non-implemented resource types
In my limited tests it works as intended and it is quite fast! Goodbye HLExtract and temporary files. 
Such a wonderful library and tool, thank you!

Usage:
`Decompiler -i "%source_vpk%" -o "decompile_extensions"` **-d -e** "vxml_c,vcss_c" 
`Decompiler -i "%source_vpk%" -o "decompile_filepath"` **-d -f** "soundstacks\\\\" 
`Decompiler -i "%source_vpk%" -o "decompile_ext_within_filepath"` **-d -e** vsndstck_c **-f** soundstacks\ 
`Decompiler -i "%source_vpk%" -o "extract_extensions"` **-e** vpcf_c,psf 
`Decompiler -i "%source_vpk%" -o "extract_one_file"` **-f** scripts/items/items_game.txt 
`Decompiler -i "%source_vpk%" -o "extract_all"`

GameTracker Usage:
`mono ../.support/SourceDecompiler/Decompiler.exe -i "$file" -o "$(echo "$file" | sed -e 's/\.vpk$/\//g')"` **-d -e** "vxml_c,vjs_c,vcss_c,vsndevts_c,vpcf_c,txt,cfg,res,png,jpg"

### DIFF
--- a/Decompiler/Options.cs
+++ b/Decompiler/Options.cs
@@ -42,6 +42,26 @@ namespace Decompiler
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
+        [Option("tool", DefaultValue = false,
+            HelpText = "Enable generic vpk dump tool, supporting filters.")]
+        public bool VPKTool { get; set; }
+
+        [Option('x', "extract",
+            HelpText = "Just extract, do not decompile")]
+        public bool Extract { get; set; }
+
+        [Option('f', "filter",
+            HelpText = "Dump filter - full or partial path, example: panorama\\")]
+        public string Filter { get; set; }
+
+        [Option('e', "filter_ext",
+            HelpText = "Dump filter - extension(s), example: vcss_c,vjs_c,vxml_c")]
+        public string FilterExt { get; set; }       
+
+        [Option('s', "silent",
+            HelpText = "Dump without printing file names")]
+        public bool Silent { get; set; }
+
         [HelpOption]
         public string GetUsage()
         {
--- a/Decompiler/Options.cs
+++ b/Decompiler/Options.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using CommandLine;
 using CommandLine.Text;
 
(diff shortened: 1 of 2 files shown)


xPaw:
I left you comments on #86 because I didn't notice it was closed, but as the code is the same they still apply.

AveYo:
I have revised the filter options, removing hard-coded extensions and the silent option.

Decompiler/Options.cs
```diff
@@ -42,6 +42,18 @@ class Options
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
+        [Option('d', "vpk_dump",
```
xPaw: I think this should be the other way around, like 'vpk_decompile', so that by default it only dumps the file.

AveYo: Well, it's called decompiler for a reason :) 
But since dumping comes first, I guess that would work too.
Will do a test and update on results.

Decompiler/Options.cs
```diff
@@ -42,6 +42,18 @@ class Options
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
+        [Option('d', "vpk_decompile", DefaultValue = false,
+            HelpText = "Decompile supported files")]
+        public bool Decompile { get; set; }
+
+        [Option('e', "vpk_extensions",
+            HelpText = "File extension(s) filter, example: vcss_c,vjs_c,vxml_c")]
+        public string ExtFilter { get; set; }
```
xPaw: CommandLine might actually support lists, try `List<string>`

AveYo: From little google that I did it's not issue-free in this older version and the library looks abandoned. 
I'm sure there is a better way to use it, but for now my implementation seems to work ok. 
Also added fix for vpk_filepath to work with any directory separator

PR #90, opened 2017-07-29
Add option to refresh cached vpk manifest - v2

AveYo:
Filtering is fine, but incomplete, since the initial cached vpk manifest will block any different output from the same source.
Currently, one needs to delete the manifest file before calling Decompiler. 
_I remember thinking about Decompiler a while ago: "such a good tool, too bad it isn't working for extracting from vpk" - and going back to HLExtract._

Decompiler would work as a generic vpk tool and skip creating the vpk manifest file - by default.

For GameTracker usage, an extra option would be added to create and reuse a cached vpk manifest file:
-c, --vpk_cache : Use cached VPK manifest

### DIFF
--- a/Decompiler/Options.cs
+++ b/Decompiler/Options.cs
@@ -44,7 +44,7 @@ namespace Decompiler
         public bool VerifyVPKChecksums { get; set; }
 
         [Option('c', "vpk_cache", DefaultValue = false,
-            HelpText = "Use cached VPK manifest")]
+            HelpText = "Use cached VPK manifest.")]
         public bool CachedManifest { get; set; }
 
         [Option('d', "vpk_decompile", DefaultValue = false,
--- a/Decompiler/Options.cs
+++ b/Decompiler/Options.cs
@@ -43,8 +43,8 @@ namespace Decompiler
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
-        [Option('c', "vpk_cache", DefaultValue = false,
-            HelpText = "Use cached VPK manifest.")]
+        [Option("vpk_cache", DefaultValue = false,
+            HelpText = "Use cached VPK manifest to keep track of updates. Only changed files will be written to disk.")]
         public bool CachedManifest { get; set; }
 
         [Option('d', "vpk_decompile", DefaultValue = false,
--- a/Decompiler/Options.cs
+++ b/Decompiler/Options.cs
@@ -43,6 +43,10 @@ namespace Decompiler
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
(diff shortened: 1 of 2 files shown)


xPaw:
Why the 2 PRs?

AveYo:
Any of the options would work for the "stand-alone" generic vpk tool functionality, but what matters the most is the root GameTracker usage. I would personally go for v2 but then I'm biased towards a generic vpk tool. What's important is to realize the need for such an option - caching is great for static usage, until it gets in your way for other tasks.

xPaw:
You also need to add an `if` around where it creates the file.

AveYo:
It's fine creating the manifest as it doubles as a logging, specially when redirecting output to null to speed up long operations. Only the sequential run with different filters being blocked was an issue.

xPaw:
It won't make much sense to make it if the option isn't used though.

AveYo:
It could be used as a starting point for a cached run if the parameters match expectations, saving a full re-dump next time, but it's fine either way. Should I refactor it to skip creating it then?

AveYo:
On second thought, it's definitely better if "stand-alone" runs would not touch the manifest, even for my specific usage (caching particles mainly, but also extracting different files). Will update accordingly.

Decompiler/Options.cs
```diff
@@ -43,6 +43,10 @@ class Options
             HelpText = "Verify checksums and signatures.")]
         public bool VerifyVPKChecksums { get; set; }
 
+        [Option('c', "vpk_cache", DefaultValue = false,
```
xPaw: Can you remove `c` and only make the full name available? This option is not useful enough to have a shorthand.

Decompiler/Options.cs
```diff
@@ -43,6 +43,10 @@ class Options
...
+            HelpText = "Use cached VPK manifest.")]
```
xPaw: A more descriptive help text wouldn't hurt.

xPaw:
Thanks!

Decompiler/Options.cs
AveYo: Done.
I was using several Decompiler calls with caching enabled in my scripts since the manifest adds up, dumping files in different folders and the shorthand was handy. Plus it was fitting into the ..,d,e,f options added - I actually went backwards from _c_ to _cache_ :)

Decompiler/Options.cs
AveYo: > Use cached VPK manifest to keep track of updates. Only changed files will be written to disk.

Can't think of anything better atm.

AveYo:
None of these were critical for your main project, I appreciate very much that you took the time to implement them just for the community.  There are a lot of people doing all sorts of modding that will love to drop manual GCFScape or hlextract usage and have an all-in-one powerful vpk tool.
Thank you!

PR #91, opened 2017-09-11
Prevent FC reading unknown resource types

AveYo:
For example, when `--vpk_decompile` a .vpk containing a game map, on non-standard .vvis_c resources 
(partially readable ones like .vrman_c are already handled)

_Unhandled Exception: System.ArgumentException: Unrecognized block type 'VXVS'
   at ValveResourceFormat.Resource.ConstructFromType(String input)
   at ValveResourceFormat.Resource.Read(Stream input)
   at Decompiler.Decompiler.DumpVPK(Package package, String type, String newType)_

Oversight on my part in the previous pull requests.
decompiler fc
decompiler

### DIFF
--- a/Decompiler/Program.cs
+++ b/Decompiler/Program.cs
@@ -654,7 +654,21 @@ namespace Decompiler
                     {
                         using (var memory = new MemoryStream(output))
                         {
-                            resource.Read(memory);
+                            try
+                            {
+                                resource.Read(memory);
+                            }
+                            catch (Exception)
+                            {
+                                lock (ConsoleWriterLock)
+                                {
+                                    Console.ForegroundColor = ConsoleColor.DarkRed;
+                                    Console.WriteLine("\tDecompiler for resource type " + type + " not implemented, extracting as-is");
+                                    Console.ResetColor();
+                                }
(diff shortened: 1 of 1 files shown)


Decompiler/Program.cs
```diff
@@ -654,7 +654,21 @@ private static void DumpVPK(Package package, string type, string newType)
...
+                            try
+                            {
+                                resource.Read(memory);
+                            }
+                            catch (Exception)
+                            {
+                                lock (ConsoleWriterLock)
+                                {
+                                    Console.ForegroundColor = ConsoleColor.DarkRed;
+                                    Console.WriteLine("\tDecompiler for resource type " + type + " not implemented, extracting as-is");
```
xPaw: You should print the actual exception instead of assuming it's not implemented.

AveYo: Well, it's descriptive enough (and having a large exception message there won't help much).

AveYo: Would this work?  
```c#
Console.WriteLine("\t" + e.Message + " on resource type " + type + ", extracting as-is");
```

PR #93, opened 2018-01-04
Port VRF and decompiler to .NET core

xPaw:
- [x] Fix Travis builds
- [x] Fix AppVeyor builds
- [ ] Make sure releases can still build an executable for decompiler
- [ ] Make sure assembly info is correct (e.g. version)

I already tried this on GameTracker-Dota2, it worked great.


.travis.yml
```diff
@@ -1,13 +1,14 @@
...
+mono: none
+dotnet: 2.0.0
 
 notifications:
   email: false
 
-install:
-  - nuget restore ValveResourceFormat.sln
+before_install:
+  - sudo apt-get install -y libgdiplus
```
yaakov-h: Why is this before install when it's literally installing something?

.travis.yml
```diff
@@ -1,13 +1,14 @@
...
 
 script:
-  - msbuild /p:Configuration=Release ValveResourceFormat.sln
-  - mono ./packages/NUnit.ConsoleRunner.3.2.0/tools/nunit3-console.exe ./Tests/bin/Release/Tests.dll
+  - dotnet build -c release -f netcoreapp2.0 ./Decompiler/Decompiler.csproj
```
yaakov-h: You might be able to get away with `msbuild /t:Restore /t:Build`, possible as one command, if not, as two.

yaakov-h: Do you really need `-f`, can this not be inferred?

Decompiler/Decompiler.csproj
```diff
@@ -1,65 +1,18 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{5F27DC08-3AE5-495C-8445-43D4E7CABA63}</ProjectGuid>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
```
yaakov-h: Use `TargetFramework` when you have a single target, only use `TargetFrameworks` when you have multiple.

Decompiler/Properties/AssemblyInfo.cs
```diff
@@ -1,11 +0,0 @@
-﻿using System.Reflection;
-
-[assembly: AssemblyTitle("Source 2 Resource Decompiler")]
-[assembly: AssemblyProduct("Source 2 Resource Decompiler")]
-[assembly: AssemblyDescription("(╯°□°）╯︵ ┻━┻")]
```
yaakov-h: R.I.P

Tests/Tests.csproj
```diff
@@ -1,75 +1,21 @@
...
-    <Optimize>true</Optimize>
-    <OutputPath>bin\Release</OutputPath>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ConsolePause>false</ConsolePause>
-    <DefineConstants>TRACE;DEBUG;</DefineConstants>
-    <DebugSymbols>true</DebugSymbols>
-    <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
-    <CodeAnalysisRuleSet>..\ValveResourceFormat\ValveResourceFormat.ruleset</CodeAnalysisRuleSet>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
```
yaakov-h: Same as above.

ValveResourceFormat/ValveResourceFormat.csproj
```diff
@@ -1,116 +1,13 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{E685E563-5F49-4D8D-8342-A7354229E54F}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>ValveResourceFormat</RootNamespace>
-    <AssemblyName>ValveResourceFormat</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
yaakov-h: Why not consume the .NET Standard library under .NET Framework? Are you trying to avoid all of the glue assemblies?

appveyor.yml
```diff
@@ -1,27 +1,33 @@
...
+image: Visual Studio 2017
 configuration: Release
+
 assembly_info:
   patch: true
   file: '**\AssemblyInfo.*'
   assembly_version: '{version}'
   assembly_file_version: '{version}'
   assembly_informational_version: '{version}'
+
```
yaakov-h: I don't believe this will patch projects using the new SDK. You might need to set the version in the csproj like so: https://github.com/WiseTechGlobal/WTG.Analyzers/blob/master/build/Global.props#L22

appveyor.yml
```diff
@@ -1,27 +1,33 @@
...
-  project: ValveResourceFormat.sln
-  verbosity: minimal
+  - nuget restore GUI\GUI.csproj
+
+build_script:
+  - dotnet build -c release
+
 after_build:
-- cmd: nuget pack ValveResourceFormat\ValveResourceFormat.csproj
+  - cmd: nuget pack ValveResourceFormat\ValveResourceFormat.csproj
```
yaakov-h: Not `dotnet pack`?

appveyor.yml
```diff
@@ -1,27 +1,33 @@
...
+
 test_script:
 - ps: >-
     $opencover = (Resolve-Path "packages/OpenCover.*/tools/OpenCover.Console.exe").ToString();
-    & $opencover -target:nunit3-console.exe "-targetargs:Tests/bin/$($env:CONFIGURATION)/Tests.dll --result=testResults.xml;format=AppVeyor" -filter:"+[ValveResourceFormat*]* -[ValveResourceFormat]SevenZip*" -output:opencoverCoverage.xml -register:user -returntargetcode
+    & $opencover "-target:%ProgramFiles%\dotnet\dotnet.exe" "-targetargs:test Tests\Tests.csproj --result=testResults.xml;format=AppVeyor" -filter:"+[ValveResourceFormat*]* -[ValveResourceFormat]SevenZip*" -output:opencoverCoverage.xml -register:user -returntargetcode
```
yaakov-h: Check that this actually works - for SK2 we get no coverage output.

.travis.yml
xPaw: Good question. https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Standard-Infrastructure

yaakov-h: Odd, but ¯\_(ツ)_/¯

.travis.yml
xPaw: I set it for VRF because so it doesn't try to build net4.6 there, so I just did the same here.

Decompiler/Decompiler.csproj
xPaw: c o n s i s t e n c y

ValveResourceFormat/ValveResourceFormat.csproj
xPaw: It won't build GUI without it due to `System.Drawing.Common`.

PR #96, opened 2018-01-08
Replace System.Drawing with SkiaSharp

xPaw:


Decompiler/Program.cs
```diff
@@ -93,12 +94,14 @@ public static void Main(string[] args)
             if (Options.CollectStats)
             {
                 Console.WriteLine();
-                Console.WriteLine("Processed resource stats:");
+                Console.WriteLine("Processed resource " +
```
xPaw: wtf did VS do here

PR #97, opened 2018-01-20
Include FBX Export and add a retry method for finding files nearby

carlosrpg:


GUI/GUI.csproj
```diff
@@ -39,6 +39,9 @@
     <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="ArcManagedFBX">
+      <HintPath>..\lib\ArcManagedFBX.dll</HintPath>
```
yaakov-h: 😢

xPaw:
Thanks for the PR and sorry for the long reply, but unfortunately I will close this PR.

I don't want to add dumping into a proprietary format, especially when it comes with a 25mb dll blob that has no license.

Issue #107, opened 2018-08-30
Incorrect image height for Dota Hero files


mdiller:
When extracting the dota vpk files, some of the images have the incorrect dimensions, causing the rest of the pixels to be filled by clamping. The following files are in the dota_vpk in `dota 2 beta/game/dota/pak01_dir.vpk`.
The following are the image locations in the vpk:
`/panorama/images/heroes/npc_dota_hero_rubick_png.vtex_c`: (dimensions should be 128 x 72)
![npc_dota_hero_rubick_png](https://user-images.githubusercontent.com/3231343/44826455-f445fd00-abc3-11e8-8e1c-fafcf68849ee.png)
`/panorama/images/heroes/selection/npc_dota_hero_rubick_png.vtex_c`: (dimensions should be 71 x 94)
![npc_dota_hero_rubick_png](https://user-images.githubusercontent.com/3231343/44826463-032caf80-abc4-11e8-99d9-11cab5891bd5.png)
The images have dimensions of (128 x 128) instead of what they should be

mdiller:
Another interesting note is that some of the files in the same directory are extracted with the correct dimensions.
`/panorama/images/heroes/selection/npc_dota_neutral_ghost_png.vtex`
![npc_dota_neutral_ghost_png](https://user-images.githubusercontent.com/3231343/44827038-1a6c9c80-abc6-11e8-992c-24672f5d5a04.png)

SinZ163:
Unfortunately for us, we haven't been able to find where/if the correct dimensions are even stored in the file, and some of these problems show up in the asset browser

AveYo:
Valve's own ` resourceinfo.exe -i test.vtex_c -game %dota% -debug tga -mip`  will have this issue.   
Ultimately a solution can be cooked up, since these _FillToPowerOfTwo_ textures can be identified, and padded bytes could be cut brute-force-ish.  
But the smart thing to do would be to use directx (or alternative?) functions to do the heavy lifting.  
D3DXCreateTextureFromFileEx / D3DXCreateTextureFromResourceEx  can return the actual image size in  D3DXIMAGE_INFO (must pass a pSrcInfo pointer when calling the function). 

working | bugged
---|---
 Block: REDI Size: `792 bytes [+0b padding]`| Block: REDI Size: `872 bytes [+0b padding]`
 Block: DATA Size: `1076 bytes [+26696b padding]`| Block: DATA Size: `1216 bytes [+16384b padding]`
[...]

PR #108, opened 2018-08-30
Output proper dimensions for FillToPowerOfTwo DXT1 / DXT5  textures

AveYo:
issue #107 PoC

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
@@ -35,9 +35,9 @@ namespace ValveResourceFormat.ResourceTypes
 
         public Dictionary<VTexExtraData, byte[]> ExtraData { get; private set; }
 
-        public ushort NonPowerOfTwoWidth { get; private set; }
+        public ushort NonPow2Width { get; private set; }
 
-        public ushort NonPowerOfTwoHeight { get; private set; }
+        public ushort NonPow2Height { get; private set; }
 
         public Texture()
         {
@@ -70,12 +70,11 @@ namespace ValveResourceFormat.ResourceTypes
             Width = reader.ReadUInt16();
             Height = reader.ReadUInt16();
             Depth = reader.ReadUInt16();
+            NonPow2Width = 0;
+            NonPow2Height = 0;
             Format = (VTexFormat)reader.ReadByte();
             NumMipLevels = reader.ReadByte();
             Picmip0Res = reader.ReadUInt32();
-
-            NonPowerOfTwoWidth = 0;
-            NonPowerOfTwoHeight = 0;
 
             var extraDataOffset = reader.ReadUInt32();
             var extraDataCount = reader.ReadUInt32();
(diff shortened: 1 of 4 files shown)


AveYo:
I'm sure there is a proper-proper way of doing it, but this is what I could come up with while being totally unfamiliar with SkiaSharp (and dotnet - I'm still using the legacy version for my projects).

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -87,6 +94,20 @@ public override void Read(BinaryReader reader, Resource resource)
 
                     reader.BaseStream.Position += offset;
 
+                    if (type == 3)
```
xPaw: Add this to `VTexExtraData` and reference it.

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -149,7 +170,7 @@ public SKBitmap GenerateBitmap()
                             }
                         }
 
-                        return DDSImage.UncompressDXT1(Reader, Width, Height);
+                        return DDSImage.UncompressDXT1(Reader, Width, Height, NonPowerOfTwoWidth, NonPowerOfTwoHeight);
```
xPaw: Does it not work if you directly replace width/height?

AveYo: No, it would return mangled bits because there is a lot of work done on them (strides colors etc).  
Someone more skilled could optimize this later on.

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
AveYo: Will do

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
AveYo: Note that all the extra work done in memory is less costly than outputting bigger padded images to disk - there are 5000+ in panorama/images, so this is actually faster overall.

mdiller:
Hey thanks for taking a look at this right away!

I tried running it with your changes and I'm getting this when I try to open the file I mentioned in the issue:
```
System.ArgumentNullException: Value cannot be null.
Parameter name: pixmap
   at SkiaSharp.SKImage.FromPixels(SKPixmap pixmap, SKImageRasterReleaseDelegate releaseProc, Object releaseContext)
   at SkiaSharp.Views.Desktop.Extensions.ToBitmap(SKBitmap skiaBitmap)
   at GUI.MainForm.ProcessFile(String fileName, Byte[] input, Package currentPackage) in C:\Development\outside_projects\temp\GUI\MainForm.cs:line 372
```
Have you tried running it on the file I mentioned, and if so are you getting this error?

AveYo:
I only checked the command line Decompiler, and it works there as I've decompiled all the DOTA images with the correct dimensions successfully.  Not so familiar with the GUI side, I was under the impression it will work there as well.  Will take a look next.

AveYo:
@mdiller : should work after this last commit, please confirm

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
AveYo: Scratch that, I gave up too soon. Fixed now (properly?)

mdiller:
Yep, seems to be working perfectly now, thanks!

Sorry for the late reply. I haven' got my github notifications configured correctly.

Issue #112, opened 2018-09-15
DXT5nm Compression


DankParrot:
Source 2 normal maps are compressed with [DXT5nm](http://wiki.polycount.com/wiki/Normal_Map_Compression#DXT5nm_Compression):

> DXT5nm is the same file format as DXT5 except before compression the red channel is moved into the alpha channel, the green channel is left as-is, and the red and blue channels are blanked with the same solid color.

When exporting a normal map, the output Red (x) channel should be the compiled texture's Alpha channel, the output Green (y) channel should stay the same and the output Blue (z) channel can be derived:
``` z = sqrt( 1 - x² - y² )```

This seems to apply to non-DXT5 normal maps too.

My guess is that the easiest way to check if a texture is a normal map would probably be to look for `NormalizeNormals` in the REDI special dependencies.

xPaw:
How did you figure out that it's DXT5nm?

DankParrot:
I tested it with a couple normal maps. My inputs were standard RGB normal maps and then the ones I extracted had no R or B channels, only G and A, and they matched up with their corresponding channels from the original map. I think I made a script at some point to try and reconstruct the third channel but it was really tedious.

If you look at a compiled normal map, they look Green and Pink instead of the usual purple-ish color of normal maps.

Example:
![copper_pot001_normal_psd_c6bb1779](https://user-images.githubusercontent.com/3376691/45600200-a09e2680-b9ad-11e8-837d-217bbc57d875.png)

xPaw:
Want to make a PR?

Perryvw:
Is it actually the responsibility of VRF to do this translation though, the preview image is the data that is actually present. Wouldn't it be up to the consuming party to transform that data however it sees fit?

DankParrot:
> Is it actually the responsibility of VRF to do this translation though, the preview image is the data that is actually present. Wouldn't it be up to the consuming party to transform that data however it sees fit?

I could understand making it an optional setting, but the functionality should be there.
Generally people would want their extracted data in a more usable form, and DXT5nm is just considered another form of compression like DXT5 normally is.

> Want to make a PR?

@AveYo made a commit to fix this at AveYo@72956f2a9c0ad3769054ee9a7aebade5b5c9658a, however I could also implement it and PR in a few minutes.

AveYo:
Just by looking at an example you can spot it is monochrome, but not gray-scale.  
[...]

PR #113, opened 2018-09-16
Perfect DXT5nm normalized normalmaps output

AveYo:
Fix for #112  

_OP test source (using the cleaned up vtex_c in my zip)_  
Imgur  
_dota_addons/overthrow/materials/models/courier/lockjaw/lockjaw_normal_  
Imgur  
_dota_addons/overthrow/materials/models/props_structures/midas_throne/midas_kobold_props_normal
Imgur_   
Last one is also `LegacySource1InvertNormals`

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
@@ -135,16 +135,20 @@ namespace ValveResourceFormat.ResourceTypes
 
                 case VTexFormat.DXT5:
                     var yCoCg = false;
+                    var normalize = false;
+                    var invert = false;
 
                     if (Resource.EditInfo.Structs.ContainsKey(ResourceEditInfo.REDIStruct.SpecialDependencies))
                     {
                         var specialDeps = (SpecialDependencies)Resource.EditInfo.Structs[ResourceEditInfo.REDIStruct.SpecialDependencies];
 
                         yCoCg = specialDeps.List.Any(dependancy => dependancy.CompilerIdentifier == "CompileTexture" && dependancy.String == "Texture Compiler Version Image YCoCg Conversion");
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/ThirdParty/DDSImage.cs
```diff
@@ -258,6 +258,13 @@ private static void DecompressBlockDXT5(int x, int y, int width, byte[] blockSto
                             finalG = ClampColor(finalAlpha + cg);
                             finalB = ClampColor(finalAlpha - co - cg);
                         }
+                        if (dxt5nm)
+                        {
+                            finalR = ClampColor(255 - finalAlpha);
```
xPaw: I don't think these need to be clamped.

AveYo: It's because of this:  
>If you treat it as YCoCg, and just replace alpha with 255 already gives a better output, but it's very bluish.
>It's the same with the formulas for CoCg_Y, so I went along with something else __since it more closely resembles tangent space normalmaps__.

xPaw: That's unrelated. What I mean is, the max value per channel is going to be `255`, so if you substract it from 255, it can't be out of 0-255 range.

AveYo: Fixed.

AveYo:
go? 👍 #112

AveYo:
It was unfortunate we worked on same code concurrently 💃 but it's all good now.  
Both FillToPowerOfTwo and DXT5nm work correctly.
How can this Travis fail be fixed?  Gonna try `<WarningsNotAsErrors>619</WarningsNotAsErrors>`

xPaw:
@DankParrot Can you test that this fixes as you expect it would?

@AveYo Perhaps all the color normalization could be done in a separate function that is called after `UncompressDXT5` instead of passing all the options into DXT5 specifically?

AveYo:
Yeah I'll do that

AveYo:
took some time - I always have floppy internet when I need it the most 💃

PR #134, opened 2019-06-01
implemented support for the new format of the bitpackedsoundinfo

### LINKED ISSUE #117: Audio type 0 is not necessarily AAC
I just found this tool and tried it out for extracting Artifact sounds. It looks like you recently added support for `audioType = 0` which was necessary for Artifact `.vsnd_c` files.

Looking at one of the Artifact sound files in `pak01_dir.vpk`, `sound/responses/set_01/centaur_warrunner/centaur_warrunner_angry_05.vsnd_c` with VRV GUI:
```
{
	uint32 m_bitpackedsoundinfo = 0x0202AC44
	int32 m_loopStart = -1
	float32 m_flDuration = 2.645125
}

Sample Rate: 2058
Bits: 17
Type: AAC
SampleSize: 6
Format: 2
Channels: 0
```

I assume the `0` at the start of `m_bitpackedsoundinfo` means audio type 0. The VRV viewer plays it just fine, but if I extract it to play it myself it looks like an MP3 rather than an AAC:

At offset 0x2F0 into the file this looks like the header of the actual file:
image

You can see "LAME 3.100" near the end of the header; suggesting LAME MP3.

If I clip the file to the screenshotted offset and load it into my favorite audio player, it shows that it's a VBR mp3:
image

Maybe just a minor nit, since operationally the tool seems to work. Thought I'd add this info.

mdiller:
Fixes #117 

I did some large scale analysis of the bitpackedsoundinfo, and found that it looks like there are 2 distinct formats. As far as I can tell, the only way to really tell the difference is that if the 5 most significant bits are all set to 0, it is definetly the new format.

I checked that it works by generating all of the files from the dota vpk, and using `ffprobe` on each of them to verify that it had a valid format.

Let me know if there are any issues with this or if anything should change.

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Sound.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Sound.cs
@@ -93,16 +93,46 @@ namespace ValveResourceFormat.ResourceTypes
 
             var bitpackedSoundInfo = ((NTROValue<uint>)Output["m_bitpackedsoundinfo"]).Value;
 
-            Type = (AudioFileType)ExtractSub(bitpackedSoundInfo, 0, 2);
-            Bits = ExtractSub(bitpackedSoundInfo, 2, 5);
-            Channels = ExtractSub(bitpackedSoundInfo, 7, 2);
-            SampleSize = ExtractSub(bitpackedSoundInfo, 9, 3);
-            AudioFormat = ExtractSub(bitpackedSoundInfo, 12, 2);
-            SampleRate = ExtractSub(bitpackedSoundInfo, 14, 17);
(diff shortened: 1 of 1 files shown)


xPaw:
Thank you!

xPaw:
Detecting 'new format' was as simple as checking the resource version itself:

https://github.com/SteamDatabase/ValveResourceFormat/blob/0bb0f9033e44621829c66719c1b955bd8cdfe341/ValveResourceFormat/Resource/ResourceTypes/Sound.cs#L70-L72

Currently digging through that code again, due to #182

mdiller:
Ahhhh that makes sense!

PR #141, opened 2019-06-18
Added various tab management options.

MattFiler:
For the sake of UX, I've added three new tab management options:

- Close all tabs to the left
- Close all tabs to the right
- Close all tabs

These options appear when right clicking on a tab in the GUI, and can also be accessed via hotkeys (with the exception of close all tabs to the left).


SinZ163:
Looks good, but with the increase in ability to close tabs, it becomes very easy to close the Console tab by accident and there is currently no way of getting that back without restarting the app, happy for that to be a seperate PR if that makes sense though.

MattFiler:
I could update this PR to make it so that the console is unable to be closed, if you'd like?

MattFiler:
The only way you can close the console is by selecting "close this tab" (right clicking on the actual console tab). I made the other options ignore the console, for that reason.

xPaw:
This seems to work fine.

> I can disable "close this tab" for the console also, if you'd prefer that?

That would be a good idea indeed.

If you want to get even fancier, the "close to left/right" options could be hidden whether the tab is first or last in the tab list.

PR #143, opened 2019-06-19
Made the tab context menu dynamic.

MattFiler:
As requested in PR #141, the context menu shown when right clicking on tabs is now dynamic.

For example, if there are no tabs to the right, the option to close tabs to the right will not be shown.

The option to close the console is now hidden as it's a little nicer for UX.


xPaw:
Thanks!

PR #152, opened 2019-06-21
Implemented file export progress dialog (#111)

r2d2rigo:
Added feature for issue #111 - a progress dialog that displays and refreshes while extracting data from the package. It is also cancelable in case the user doesn't want to wait for it to finish.

### DIFF
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -34,7 +34,8 @@ namespace GUI.Forms
         protected override void OnShown(EventArgs e)
         {
             Task
-                .Run(async () =>
+                .Run(
+                async () =>
                 {
                     Invoke((Action)(() =>
                     {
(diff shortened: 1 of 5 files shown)


GUI/Forms/ExtractProgressForm.cs
```diff
@@ -0,0 +1,117 @@
...
+                    await ExtractFilesAsync();
+                });
+        }
+
+        protected override void OnClosing(CancelEventArgs e)
+        {
+            cancellationTokenSource.Cancel();
+        }
+
+        private async Task CalculateFilesToExtractAsync(TreeNode root)
```
xPaw: This is async but doesn't actually do any async work?

xPaw:
Thanks, this looks great. One thing though, either the progress window should be closed on end, or the text should be changed to say its finished.

GUI/Forms/ExtractProgressForm.cs
r2d2rigo: Sorry, the habit of always making everything async :)

r2d2rigo:
Done!

PR #162, opened 2019-06-26
Fix (loose) panorama files not being decompiled properly

13k:
When decompiling loose Panorama files (not in a VPK), the appropriate `ResourceType`s weren't matched accordingly.

It also handles the file extension special case for layout, script and style. Not sure if it's even valid to do that for `ResourceType.Panorama`, since it seems to mean "unknown Panorama resource type". Resources with this type do even pass the [extension check](https://github.com/SteamDatabase/ValveResourceFormat/blob/d2a3d69323f8049e150d0f05132757a1c7c9a535/Decompiler/Decompiler.cs#L237-L246)?

### DIFF
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -327,6 +327,14 @@ namespace Decompiler
                             data = ((Panorama)resource.Blocks[BlockType.DATA]).Data;
                             break;
 
+                        case ResourceType.PanoramaLayout:
+                        case ResourceType.PanoramaScript:
+                        case ResourceType.PanoramaStyle:
+                            // .vcss -> .css, .vjs -> .js, .vxml -> .xml
+                            extension = "." + extension.Substring(2, extension.Length - 2);
+                            data = ((Panorama)resource.Blocks[BlockType.DATA]).Data;
+                            break;
+
                         case ResourceType.Sound:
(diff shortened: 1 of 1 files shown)


Decompiler/Decompiler.cs
```diff
@@ -327,6 +327,14 @@ private void ProcessFile(string path, Stream stream)
                             data = ((Panorama)resource.Blocks[BlockType.DATA]).Data;
                             break;
 
+                        case ResourceType.PanoramaLayout:
+                        case ResourceType.PanoramaScript:
+                        case ResourceType.PanoramaStyle:
+                            // .vcss -> .css, .vjs -> .js, .vxml -> .xml
+                            extension = "." + extension.Substring(2, extension.Length - 2);
```
xPaw:
```suggestion
                            extension = extension.Substring(1);
```

Shouldn't it be like this?

xPaw:
Ideally standalone and files and VPK dumping would use the case code path, so the dumping code doesn't have to be duplicated.

xPaw:
I made some changes in CLI to combine vpk/non-vpk dumping, can you test?

https://github.com/SteamDatabase/ValveResourceFormat/compare/file-extract

13k:
Sorry to get back to you so late. It seems to be working. Nice refactoring too. Thanks!

PR #165, opened 2019-06-27
Consolidate file dumping/exporting into VRF helper function

### LINKED ISSUE #164: Consolidate file dumping/exporting into a single place
There's three separate code paths responsible for dumping files to disk:
1. File passed to cli directly
2. Files found in vpk with cli
3. Export files in GUI

They all do similiar things like getting the file extension and calling the relevant export/tostring method based on resource type.

I'm thinking a separate class can be created in Decompiler project which would be responsible for dumping, and GUI can reference and use it too.

xPaw:
Fixes #164

There's two dodgy things that I'm not entirely sure about:
- Extension handling in CLI, there's some magic involved
- `default` resource type tostring handling

Not sure what we should do about either of these.

### DIFF
--- a/GUI/MainForm.cs
+++ b/GUI/MainForm.cs
@@ -1068,7 +1068,7 @@ namespace GUI
                     extensions = new[] { ((Sound)resource.Blocks[BlockType.DATA]).Type.ToString().ToLower() };
                     break;
                 case ResourceType.Texture:
-                    extensions = new[] { "png", "jpg", "gif", "bmp" };
+                    extensions = new[] { "png" };
                     break;
                 case ResourceType.PanoramaLayout:
                     extensions = new[] { "xml", "vxml" };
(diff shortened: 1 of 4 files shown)


GUI/MainForm.cs
```diff
@@ -1068,7 +1068,7 @@ private void ExportToolStripMenuItem_Click(object sender, EventArgs e)
                     extensions = new[] { ((Sound)resource.Blocks[BlockType.DATA]).Type.ToString().ToLower() };
                     break;
                 case ResourceType.Texture:
-                    extensions = new[] { "png", "jpg", "gif", "bmp" };
+                    extensions = new[] { "png" };
```
SinZ163: Losing JPG and gif support?

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -0,0 +1,94 @@
...
+                    break;
+
+                // These all just use ToString() and WriteText() to do the job
+                case ResourceType.SoundEventScript:
+                    data = Encoding.UTF8.GetBytes(resource.Blocks[BlockType.DATA].ToString());
+                    break;
+
+                default:
+                    Console.WriteLine("-- (I don't know how to dump this resource type)"); // TODO: What do we do with this
+                    data = Encoding.UTF8.GetBytes(resource.Blocks[BlockType.DATA].ToString());
```
SinZ163: Switching from not dumping on decompiler to dumping potentially an error is a change in behaviour, is this what we want?

GUI/MainForm.cs
xPaw: Yes. I thought about it and png is the most sane format for this (and other code paths didnt support other formats).

ValveResourceFormat/IO/FileExtract.cs
xPaw: That's what I mentioned in the initial post, different code paths handled this differently, and I'm not sure how we want to handle it here.

xPaw: When decompiling from vpk here: https://github.com/SteamDatabase/ValveResourceFormat/blob/41f467ca61cc0fe78b10f01522bbaef25ba71703/Decompiler/Decompiler.cs#L660

it will fallback to dumping the file as-is when an exception is thrown. Other code paths would error out (if DATA block is missing, or it has no ToString), so we should be fine?

PR #206, opened 2020-03-24
Texture fixes

lewa-j:
The BC7 had issues with alpha.
Before:
img
After:
img
The BC6H looked wrong on especially bright spots, such as the sun.
Also cubemap textures (skybox and envmaps) had incorrect offset for mips.
Before:
img 
img
After:
img
img

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
@@ -449,7 +449,7 @@ namespace ValveResourceFormat.ResourceTypes
                 }
                 else
                 {
-                    offset = CalculateBufferSizeForMipLevel(j);
+                    offset = CalculateBufferSizeForMipLevel(j) * (Flags.HasFlag(VTexFlags.CUBE_TEXTURE) ? 6 : 1);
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -449,7 +449,7 @@ private void SkipMipmaps()
                 }
                 else
                 {
-                    offset = CalculateBufferSizeForMipLevel(j);
+                    offset = CalculateBufferSizeForMipLevel(j) * (Flags.HasFlag(VTexFlags.CUBE_TEXTURE) ? 6 : 1);
```
xPaw: Interesting. Does this work for other encodings like dxt5?

lewa-j: Yes. This envmap on my shots is DXT5.

xPaw: Okay, I tested it on Dota, and it does produce a better texture!

xPaw:
Thank you!

PR #207, opened 2020-03-24
More textures work

lewa-j:
Fix ATI1N
Add support for ATI2N
Refactor DXT5


xPaw:
Looks like something broke in DXT5, one of the tests is failing.

lewa-j:
Oh, i see. I for got about NPOT textures.

xPaw:
Thanks! Just ran `--stats` on both HL:A and Dota 2, no crashes showed up.

Getting pretty close to actually supporting all the formats in the end 😊

PR #217, opened 2020-03-28
Textures, materials and shaders improvements

lewa-j:
Should fix https://github.com/SteamDatabase/ValveResourceFormat/issues/114 for viewer, but not sure about export.
before:
img
after:
img
before:
img
after:
img
before:
vrf exe Screenshot 2020 03 29 - 00 36 59 94
after:
vrf exe Screenshot 2020 03 28 - 17 15 12 48

### DIFF
--- a/GUI/Types/Renderer/Shaders/simple.frag
+++ b/GUI/Types/Renderer/Shaders/simple.frag
@@ -62,8 +62,8 @@ void main()
 #else
     //Calculate half-lambert lighting
     float illumination = dot(worldNormal, lightDirection);
-    illumination = illumination * 0.5 + 0.5;
-    illumination = illumination * illumination;
+    //illumination = illumination * 0.5 + 0.5;
+    //illumination = illumination * illumination;
     illumination = min(illumination + 0.3, 1.0);
 #endif
 
@@ -86,14 +86,14 @@ void main()
 #endif
 
 #if param_renderMode_Normals
-	outputColor = vec4(vNormalOut, 1.0);
+	outputColor = vec4(vNormalOut * vec3(0.5) + vec3(0.5), 1.0);
 #endif
 
 #if param_renderMode_BumpNormals
-	outputColor = vec4(vNormalOut, 1.0);
+	outputColor = vec4(worldNormal * vec3(0.5) + vec3(0.5), 1.0);
 #endif
 
 #if param_renderMode_Illumination == 1
-	outputColor = vec4(illumination, 0.0, 0.0, 1.0);
+	outputColor = vec4(illumination, illumination, illumination, 1.0);
 #endif
 }
--- a/GUI/Types/Renderer/Shaders/simple.frag
+++ b/GUI/Types/Renderer/Shaders/simple.frag
@@ -60,12 +60,9 @@ void main()
 #if param_renderMode_FullBright == 1 || param_F_FULLBRIGHT == 1
     float illumination = 1.0;
 #else
-    //Calculate half-lambert lighting
+    //Calculate lambert lighting
     float illumination = max(0.0, dot(worldNormal, lightDirection));
-    //illumination = illumination * 0.5 + 0.5;
-    //illumination = illumination * illumination;
-    //illumination = min(illumination + 0.3, 1.0);
-    illumination = illumination * 0.7 + 0.3;
+    illumination = illumination * 0.7 + 0.3;//add ambient
 #endif
 
     //Calculate tint color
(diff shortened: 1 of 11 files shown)


lewa-j:
Wait, don't merge. I realized that changes in ShaderLoader were a mistake.

lewa-j:
Now good.

xPaw:
hemiOctRB is likely used in DXT5 too, did you look into that?

GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -62,8 +62,8 @@ void main()
 #else
     //Calculate half-lambert lighting
     float illumination = dot(worldNormal, lightDirection);
-    illumination = illumination * 0.5 + 0.5;
-    illumination = illumination * illumination;
+    //illumination = illumination * 0.5 + 0.5;
+    //illumination = illumination * illumination;
```
Perryvw: This should be reverted, this formula is the half-lambert equation Valve is known to use.

GUI/Types/Renderer/Shaders/vr_standard.frag
```diff
@@ -74,9 +94,10 @@ void main()
 #else
     //Calculate half-lambert lighting
     float illumination = dot(worldNormal, lightDirection);
-    illumination = illumination * 0.5 + 0.5;
-    illumination = illumination * illumination;
-    illumination = min(illumination + 0.3, 1.0);
+    //illumination = illumination * 0.5 + 0.5;
+    //illumination = illumination * illumination;
+    //illumination = min(illumination + 0.3, 1.0);
+    illumination = illumination * 0.7 + 0.3;
```
Perryvw: This should probably also be reverted

ValveResourceFormat/Resource/ResourceTypes/Material.cs
```diff
@@ -93,6 +96,13 @@ public override void Read(BinaryReader reader, Resource resource)
                 arguments.Add(name, value != 0);
             }
 
+            var specialDeps = (SpecialDependencies)Resource.EditInfo.Structs[ResourceEditInfo.REDIStruct.SpecialDependencies];
```
Perryvw: do we really have to use editinfo for this? Is there not some flag in the material DATA that we can use instead (they are automatically loaded into the shader)

lewa-j:
I don't found any dxt5 normalmaps with it. Do you know examples in hla or steamvr?

GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -86,14 +86,14 @@ void main()
 #endif
 
 #if param_renderMode_Normals
-	outputColor = vec4(vNormalOut, 1.0);
+	outputColor = vec4(vNormalOut * vec3(0.5) + vec3(0.5), 1.0);
```
Perryvw: All fragment shaders should visualize normals and bump normals the same. So either revert this change or update the renderMode code in the other fragment shaders too. I personally prefer keeping it the way it was.

GUI/Types/Renderer/Shaders/simple.frag
lewa-j: I found mentions only about Source 1. Source 2 should have a more physically accurate lighting model.

ValveResourceFormat/Resource/ResourceTypes/Material.cs
lewa-j: I don't see any. At least from that VRF shows.

GUI/Types/Renderer/Shaders/simple.frag
lewa-j: As it was before, all negative values would be just black. It half the spectrum. I'll update the rest.

GUI/Types/Renderer/Shaders/simple.frag
Perryvw: Unless we find explicit references for this we assume Source 2 is doing the same as Source 1.

lewa-j: http://media.steampowered.com/apps/valve/2015/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf
page 42

Perryvw: Very interesting, in that case can you remove the old commented code?

lewa-j: Sure

PR #223, opened 2020-03-29
Replace uses of FolderBrowserDialog with Ookii.Dialogs' VistaFolderBrowserDialog

JillCrungus:
The "VistaFolderBrowserDialog" provided by the Ookii.Dialogs library is a lot nicer and easier to use than the default FolderBrowserDialog. It looks a lot more like the file browser dialog and doesn't require the user to manually navigate through the directory tree in such a tedious way.

### DIFF
--- a/GUI/Forms/SettingsForm.cs
+++ b/GUI/Forms/SettingsForm.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using GUI.Utils;
+using Ookii.Dialogs.WinForms;
 
 namespace GUI.Forms
 {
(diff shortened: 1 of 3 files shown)


xPaw:
Looks fine to me.

Shall we also use `Ookii.Dialogs.WinForms.ProgressDialog` for extraction progress bar?

JillCrungus:
> Looks fine to me.
> 
> Shall we also use `Ookii.Dialogs.WinForms.ProgressDialog` for extraction progress bar?
> 
> EDIT: It looks like winforms on github has new dialog: [dotnet/winforms#57](https://github.com/dotnet/winforms/pull/57)
> I think if we port to .net core (https://github.com/SteamDatabase/ValveResourceFormat/tree/netcoreapp3-winforms), it will use the new dialog without using third party library?

I don't know if the other components of Ookii are worth using because I'm not well versed on what advantages they provide, the advantage of the folder browser is obvious though.

I didn't know about that Winforms change - definitely worth looking into I think if it means not needing to depend on the third party lib.

xPaw:
Okay I merged the .net core winforms port, so this is no longer needed.

PR #227, opened 2020-03-30
HemiOct impovements

lewa-j:
After examining disassembly of HLA shaders using RenderDoc, i found out that valve used a simplified formula, and not the one that was in the [article](http://jcgt.org/published/0003/02/01/).

Additionally, found a BC7 normal map with an inverted Y component. (materials/models/props_lab/server_09_normal_tga_3881d0a5.vtex_c) Handled that too

BTW RenderDoc is incredible
hlvrcapture

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/TextureDecompressors.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/TextureDecompressors.cs
@@ -505,20 +505,9 @@ namespace ValveResourceFormat.ResourceTypes
                             {
                                 if (hemiOct)
                                 {
-                                    float SignNotZero(float v)
-                                    {
-                                        return (v >= 0.0f) ? +1.0f : -1.0f;
-                                    }
-
-                                    float nx = ((data[dataIndex + 3] / 255.0f) * 2) - 1;
-                                    float ny = ((data[dataIndex + 1] / 255.0f) * 2) - 1;
+                                    float nx = ((data[dataIndex + 3] + data[dataIndex + 1]) / 255.0f) - 1.003922f;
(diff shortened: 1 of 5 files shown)


ValveResourceFormat/Resource/ResourceTypes/TextureDecompressors.cs
```diff
@@ -505,20 +505,9 @@ public static void UncompressDXT5(SKImageInfo imageInfo, BinaryReader r, Span<by
...
                                 if (hemiOct)
                                 {
-                                    float SignNotZero(float v)
-                                    {
-                                        return (v >= 0.0f) ? +1.0f : -1.0f;
-                                    }
-
-                                    float nx = ((data[dataIndex + 3] / 255.0f) * 2) - 1;
-                                    float ny = ((data[dataIndex + 1] / 255.0f) * 2) - 1;
+                                    float nx = ((data[dataIndex + 3] + data[dataIndex + 1]) / 255.0f) - 1.003922f;
```
xPaw: Any idea what's the magic behind `1.003922`?

xPaw:
Can you fix up identation in the shader please.

lewa-j:
Done

ValveResourceFormat/Resource/ResourceTypes/TextureDecompressors.cs
lewa-j: It was in the disassembly. I think this is to avoid a zero length vector.

xPaw:
Thanks.

Maybe you should join moddota discord :)

lewa-j:
Not a big fan of dota games. But since there is a source 2 reverse community, i probably should.

xPaw:
I don't play Dota either, but the VRF devs are there, so 🤷‍♂

PR #234, opened 2020-04-06
GLTF Initial Texture Import

Lonswaya:
This is just basic texture exporting and gltf links to the images that we export alongside the file

TODO
- Texture Transparency
- Reading vmat for texture types (instead of filename)
- Cheaper method for replacing alpha on images
- Exporting to TGA instead of PNG? At least that's what valve does
- Set exported image URIs to reflect actual names

### DIFF
--- a/GUI/Types/Exporter/GltfModelExporter.cs
+++ b/GUI/Types/Exporter/GltfModelExporter.cs
@@ -151,9 +151,9 @@ namespace GUI.Types.Exporter
         private Material GenerateGLTFMaterialFromRenderMaterial(RenderMaterial renderMaterial, ModelRoot model, VrfGuiContext context, string materialName)
         {
             var material = model
-                    .CreateMaterial(materialName)
-                    .WithPBRMetallicRoughness();
-            material.Alpha = AlphaMode.MASK;
+                    .CreateMaterial(materialName);
+
+            material.Alpha = renderMaterial.IsBlended ? AlphaMode.BLEND : AlphaMode.OPAQUE;
 
             foreach (var renderTexture in renderMaterial.Material.TextureParams)
             {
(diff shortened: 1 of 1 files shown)


GUI/Types/Exporter/GltfModelExporter.cs
```diff
@@ -43,11 +47,12 @@ public void ExportToFile(string resourceName, string fileName, VModel model, Vrf
                 {
                     continue;
                 }
-
-                var name = Path.GetFileName(meshReference);
+                var nodeName = Path.GetFileName(meshReference);
```
Perryvw, line 50: GetFileNameWithoutExtension might be nicer here (they are all vmesh anyway)

GUI/Types/Exporter/GltfModelExporter.cs
```diff
@@ -130,16 +136,102 @@ private Mesh CreateGltfMesh(string meshName, VMesh vmesh, ModelRoot model)
...
                 }
             }
 
             return mesh;
         }
 
+        private Material GenerateGLTFMaterialFromRenderMaterial(RenderMaterial renderMaterial, ModelRoot model, VrfGuiContext context, string materialName)
+        {
+            var material = model
+                    .CreateMaterial(materialName + "_material")
```
Perryvw, line 155: Is this _material postfix really needed, isn't it obvious this is a material?

GUI/Types/Exporter/GltfModelExporter.cs
```diff
@@ -130,16 +136,102 @@ private Mesh CreateGltfMesh(string meshName, VMesh vmesh, ModelRoot model)
...
+                            var pixelAt = bitmap.GetPixel(row, col);
+                            bitmap.SetPixel(col, row, new SKColor(pixelAt.Red, pixelAt.Green, pixelAt.Blue,  255));
+                        }
+                    }
+                }
+                var textureImage = SKImage.FromBitmap(bitmap);
+                using var data = textureImage.Encode(SKEncodedImageFormat.Png, 100);
+
+                var image = model.UseImageWithContent(data.ToArray());
+                // TODO find a way to change the image's URI to be the image name, right now it turns into (model)_0, (model)_1....
```
Perryvw, line 189: Is this comment still accurate? Seems okay to use the fileName to me, it might just happen to have the same name as the model ofcourse...

Lonswaya: You can't rename the image's uri, so for a file gordon.gltf the associated images will be gordon_0.png, gordon_1.png... etc

This will likely be fixed by writing a post-export script

GUI/Types/Exporter/GltfModelExporter.cs
Lonswaya: I assumed names had to be unique, but you are right. I'll change this for all materials, textures, images, and samplers

xPaw:
If you right click in vpk and export from there, material loader will fail because it has OpenGL calls in there.

Lonswaya:
> 
> 
> If you right click in vpk and export from there, material loader will fail because it has OpenGL calls in there.

This bug would have still existed if the "Export to OBJ" logic was used instead of exporting to GLTF, see my comment in the #steamdb-vrf channel for more context

xPaw:
On master, it didn't export to obj from right click as it was a separate code path.

PR #246, opened 2020-05-09
More entity keys

KonqiTheKonqueror:


ValveResourceFormat/Utils/EntityLumpKnownKeys.cs
```diff
@@ -1904,6 +1925,7 @@ public EntityLumpKnownKeys()
                 "particletype",
                 "passthroughcaller",
                 "pathcornername",
+                "pathNodes",
```
xPaw, line 1928: uppercase

KonqiTheKonqueror: I corrected this in the second commit, sorry about that!

PR #270, opened 2020-07-19
Cap GLViewerControl FPS to 500

WYVERN2742:
In some cases, `stopwatch.ElapsedMilliseconds` could be 0 when a frame is rendered in less than a milisecond. This caused issues with advancing to the next frame and stuttering.


xPaw:
@Perryvw Do we have a better fix for this?

Issue #309, opened 2020-11-19
value error on export of model 


CrazyHairGuy:
### Describe the bug
I get "value" error on exporting a model as glTF.

### Reference file in a Valve game
\Half-Life Alyx\game\hlvr\pak01_dir.vpk\models\creatures\headcrab_classic\headcrab_classic.vmdl_c

### Expected behavior
An exported glTF file

### Version info
VRF v0.1.6.1397, Windows 10 ver. 1909

### Additional context
Here's the console output:

``` ---> System.NotFiniteNumberException: value
   at SharpGLTF.Memory.FloatingAccessor.set_Item(Int32 rowIndex, Int32 subIndex, Single value)
   at SharpGLTF.Memory.EncodedArrayUtils._CopyTo[T](IEnumerable`1 src, IList`1 dst, Int32 dstOffset)
   at SharpGLTF.Schema2.AnimationSampler._CreateOutputAccessor(IReadOnlyList`1 output)
   at SharpGLTF.Schema2.AnimationSampler.SetKeys(IReadOnlyDictionary`2 keyframes)
   at ValveResourceFormat.IO.GltfModelExporter.<ExportToFile>g__AddMeshNode|14_0(String name, Mesh mesh, Skeleton skeleton, <>c__DisplayClass14_0& ) in C:\projects\valveresourceformat\ValveResourceFormat\IO\GltfModelExporter.cs:line 101
   at ValveResourceFormat.IO.GltfModelExporter.ExportToFile(String resourceName, String fileName, Model model) in C:\projects\valveresourceformat\ValveResourceFormat\IO\GltfModelExporter.cs:line 121
   at GUI.Types.Exporter.ExportFile.<>c__DisplayClass1_0.<Export>b__0(Object _, EventArgs __) in C:\projects\valveresourceformat\GUI\Types\Exporter\ExportFile.cs:line 88
   at GUI.Forms.GenericProgressForm.<OnShown>b__6_0() in C:\projects\valveresourceformat\GUI\Forms\GenericProgressForm.cs:line 37
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__274_0(Object obj)
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
[...]

Issue #232, opened 2020-04-03
Implement exporting models to glTF


xPaw:
https://github.com/vpenades/SharpGLTF looks like a good lib to use.

Some thoughts are here: https://docs.google.com/document/d/12QWpRG5hnRU4pw0XLuWLj_SGH7BlaqAWIwqj0mhnYd4/view

xPaw:
Work is being done here: https://github.com/SteamDatabase/ValveResourceFormat/tree/gltf

Issue #208, opened 2020-03-25
Optimize file loading in gui (map viewer)


xPaw:
When a big map loads, it keeps a lot of cached resources around, which consumes entirely too much memory in HL:A maps.

xPaw:
Animations (loaded from entities) seem to be one of the major contributors to this.

PR #344, opened 2021-04-19
gltf exporter - Added support for WorldNode and World

vermadas:
Found this project while looking for a way to bring Valve's SteamVR environments into Blender.

Only being able to export one model at a time was too restrictive, so I added the ability to export both vwrld (World) and vwnod (WorldNode) files to glTF in the GUI. Related issue: #232 

This was tested against the default SteamVR environment and a few others from the workshop. I attempted the first part of the intro level to Alyx but it failed when it tried to export a Strider and ran into #309 

As you might expect, more complex environments will take a long time and consume lots of memory. Related to #208, I think the app is not properly disposing of loaded resources, and that is the primary reason for the high memory consumption.


ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -32,107 +36,260 @@ public class GltfModelExporter
...
+            {
+                var modelName = entity.GetProperty<string>("model");
+                if (string.IsNullOrEmpty(modelName))
+                {
+                    // Only worrying about models for now
+                    continue;
+                    // TODO: Think about adding lights with KHR_lights_punctual
+                }
+                var className = entity.GetProperty<string>("classname");
+                if (className == "vr_teleport_area" || className == "trigger_multiple")
```
xPaw, line 109: Not the biggest fan of doing a filter on entities here, there's gonna be more different triggers too.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -32,107 +36,260 @@ public class GltfModelExporter
...
+            var pitchYawRoll = ParseVector(angles);
+            var rollMatrix = Matrix4x4.CreateRotationX(pitchYawRoll.Z * ((float)Math.PI / 180f)); // Roll
+            var pitchMatrix = Matrix4x4.CreateRotationY(pitchYawRoll.X * ((float)Math.PI / 180f)); // Pitch
+            var yawMatrix = Matrix4x4.CreateRotationZ(pitchYawRoll.Y * ((float)Math.PI / 180f)); // Yaw
+
+            var rotationMatrix = rollMatrix * pitchMatrix * yawMatrix;
+            return scaleMatrix * rotationMatrix * positionMatrix;
+        }
+
+        private static Vector3 ParseVector(string input)
```
xPaw, line 196: Did this and the function above get copied from the renderer? Would be better to split them into a separate class instead of duplcating the code.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -185,24 +342,116 @@ public void ExportToFile(string resourceName, string fileName, VMesh mesh)
...
+            VMesh mesh, Skeleton skeleton, string skinMaterialPath = null)
+        {
+            if (mesh.GetData().GetArray("m_sceneObjects").Length == 0)
+            {
+                return null;
+            }
+
+            var hasJoints = skeleton != null && skeleton.AnimationTextureSize > 0;
+
+            if (MeshNameCounterDict.TryGetValue(name, out var counter))
```
xPaw, line 368: What problem is this solving exactly?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -185,24 +342,116 @@ public void ExportToFile(string resourceName, string fileName, VMesh mesh)
...
+            {
+                // Give it a unique name regardless
+                var existingNode = scene.FindNode(n => n.Name == name);
+                name = $"{name}.{counter++:000}";
```
xPaw, line 372: Does this counter increment even update in `MeshNameCounterDict`?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -503,24 +767,25 @@ private Material GenerateGLTFMaterialFromRenderMaterial(VMaterial renderMaterial
...
+                        for (var i = 0; i < bitmapSpan.Length; i++)
+                        {
+                            bitmapSpan[i] = bitmapSpan[i].WithAlpha(255);
+                        }
                     }
-                }
 
-                string exportedTexturePath = Path.Join(DstDir, fileName);
-                exportedTexturePath = Path.ChangeExtension(exportedTexturePath, "png");
+                    exportedTexturePath = Path.Join(DstDir, fileName);
```
xPaw, line 785: You can move these manipulations outside of the `using`.

xPaw (commented):
Thanks for the PR! I did a brief pass and left comments on what jumped out to me. Still needs a little more careful review and testing.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -32,107 +36,260 @@ public class GltfModelExporter
...
+                var childEntityLump = (VEntityLump)childEntityLumpResource.DataBlock;
+                LoadEntityMeshes(exportedModel, scene, childEntityLump);
+            }
+        }
+
+        private static string GetSkinPathFromModel(VModel model, string skinName)
+        {
+            var materialGroupForSkin = model.Data.GetArray<IKeyValueCollection>("m_materialGroups")
+                .ToList()
+                .Where(m => m.GetProperty<string>("m_name") == skinName)
```
xPaw, line 161: You can pass this filter into `SingleOrDefault` directly.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -32,107 +36,260 @@ public class GltfModelExporter
...
+                .SingleOrDefault();
+
+            if (materialGroupForSkin == null) return null;
```
xPaw, line 164: There's a few places where you do this, but please use brackets like so:

```csharp
if (materialGroupForSkin == null)
{
    return null;
}
```

*Need to add editorconfig so this is enforced*

GUI/Types/Exporter/ExportFile.cs
```diff
@@ -11,6 +12,9 @@ namespace GUI.Types.Exporter
 {
     public static class ExportFile
     {
+        static ISet<ResourceType> ResourceTypesThatAreGltfExportable = new HashSet<ResourceType>()
```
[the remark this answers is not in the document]
vermadas: It's referenced in two places. Here:
https://github.com/vermadas/ValveResourceFormat/blob/4b93847b3c051f3ed38bd2d8d0e627be5b54fbb4/GUI/Types/Exporter/ExportFile.cs#L31
And here:
https://github.com/vermadas/ValveResourceFormat/blob/4b93847b3c051f3ed38bd2d8d0e627be5b54fbb4/GUI/Types/Exporter/ExportFile.cs#L69
It helps in the second instance to not have the exporter declaration for every case. Seemed better than having a large if condition.

ValveResourceFormat/IO/GltfModelExporter.cs
vermadas: I'll take that out.

ValveResourceFormat/IO/GltfModelExporter.cs
vermadas: It did. I can move those into a separate class in the ValveResourceFormat project.

ValveResourceFormat/IO/GltfModelExporter.cs
vermadas: If multiple instances of the same mesh exist, instead of exporting a copy of the mesh in the glTF each time, export it once and then add references to it for the duplicates.

ValveResourceFormat/IO/GltfModelExporter.cs
vermadas: You're right, not sure how I missed that one.

ValveResourceFormat/IO/GltfModelExporter.cs
vermadas: The last line in the using, bitmap.PeekPixels().Encode(), uses a FileStream which takes the exportedTexturePath in the constructor.

vermadas: Ah right, will fix.

ValveResourceFormat/IO/GltfModelExporter.cs
Perryvw: Proposal to improve this: Get rid of the counter and just store a name -> node mapping in the dictionary. Then this if statement just becomes:
```cs
if (LoadedMeshDictionary.TryGetValue(name, out var existingNode)) {
    if (!hasJoints && existingNode.Skin == null)
    {
        var newNode = scene.CreateNode(name);
        newNode.Mesh = existingNode.Mesh;
        return newNode;
    }
}
 ```

vermadas: Usage of the counter was to keep the node names unique, but the glTF spec doesn't seem to care, so I went ahead with this change. Thanks!

Perryvw (approved):

xPaw (approved):

PR #350, opened 2021-05-01
Vphys viewer

lewa-j:
Ability to redner vphys_c files and embeded in vmdl_c colliders.
Checkboxes to show all triggers/entity colliders in world viewer.

img
img
img

### DIFF
--- a/GUI/Types/Renderer/GLModelViewer.cs
+++ b/GUI/Types/Renderer/GLModelViewer.cs
@@ -138,6 +138,11 @@ namespace GUI.Types.Renderer
             {
                 physSceneNode = new PhysSceneNode(Scene, phys);
                 Scene.Add(physSceneNode, false);
+
+                //disabled by dafault. Enable if viewing only phys
+                physSceneNode.enabled = (model == null);
+
+                ViewerControl.AddCheckBox("Show Physics", physSceneNode.enabled, (v) => { physSceneNode.enabled = v; });
             }
         }
 
--- a/GUI/Types/Renderer/GLModelViewer.cs
+++ b/GUI/Types/Renderer/GLModelViewer.cs
@@ -78,6 +78,19 @@ namespace GUI.Types.Renderer
                 Scene.Add(modelSceneNode, false);
 
                 phys = model.GetEmbeddedPhys();
+                if (phys == null)
+                {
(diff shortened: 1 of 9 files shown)


GUI/Types/Renderer/GLModelViewer.cs
```diff
@@ -69,6 +77,21 @@ protected override void LoadScene()
                 SetAvailableAnimations(modelSceneNode.GetSupportedAnimationNames());
                 Scene.Add(modelSceneNode, false);
 
+                phys = model.GetEmbeddedPhys();
+                if (phys == null)
+                {
+                    var refPhysicsPaths = model.GetReferencedPhysNames();
+                    if (refPhysicsPaths.Any())
+                    {
+                        //TODO are there any models with more than one vphys?
```
xPaw, line 86: Add a log for this so we can at least catch it for the future?

GUI/Types/Renderer/GLSceneViewer.cs
```diff
@@ -87,7 +87,8 @@ private void OnLoad(object sender, EventArgs e)
             if (Scene.AllNodes.Any())
             {
                 var bbox = Scene.AllNodes.First().BoundingBox;
-                var location = new Vector3(bbox.Max.Z, 0, bbox.Max.Z) * 1.5f;
+                //if first node has no bbox, LookAt will break camera, so +1 to location.x
+                var location = new Vector3(bbox.Max.Z + 1, 0, bbox.Max.Z) * 1.5f;
```
xPaw, line 91: Shouldn't the second Z also need +1?

GUI/Types/Renderer/WorldLoader.cs
```diff
@@ -147,6 +147,9 @@ private void LoadEntitiesFromLump(Scene scene, LoadResult result, EntityLump ent
                     classname == "sky_camera" ||
                     classname == "point_devshot_camera" ||
                     classname == "point_camera";
+                var isTrigger =
+                    classname.Contains("trigger") ||
```
xPaw, line 151: StartsWith is probably better unless you know of a trigger that doesn't?

GUI/Types/Renderer/GLSceneViewer.cs
lewa-j: Not really. We just need any non zero length vector

PR #370, opened 2021-08-11
Parsing dynamic expression in Material export

robert-hoff:


ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+            ("min",        2),     // 1C
+            ("max",        2),     // 1D
+            ("SrgbLinearToGamma",1), // 1E
+            ("SrgbGammaToLinear",1), // 1F
+            ("random",     2),     // 20
+            ("normalize",  1),     // 21
+            ("length",     1),     // 22
+            ("sqr",        1),     // 23
+            ("TextureSize",1),     // 24
+        };
```
xPaw, line 59: There's 3 more in dota: rotation2d, rotate2d, sincos

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+
+        private static readonly string[] OperatorSymbols = {
+            "","","","","","","","","","","","","",
+            "==","!=",">",">=","<","<=","+","-","*","/","%"};
+
+        private enum OPCODE
```
xPaw, line 65: Can you please add `// unknown = 0...` in the gaps where we don't know what opcode is.

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+            ("SrgbGammaToLinear",1), // 1F
+            ("random",     2),     // 20
+            ("normalize",  1),     // 21
+            ("length",     1),     // 22
+            ("sqr",        1),     // 23
+            ("TextureSize",1),     // 24
+        };
+
+        private static readonly string[] OperatorSymbols = {
+            "","","","","","","","","","","","","",
```
xPaw, line 62: What's with the empty?

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+                i--;
+            }
+            return swizzle.Substring(0, i + 1);
+        }
+
+        private static string Trimb(string exp)
+        {
+            return exp[0] == '(' && exp[^1] == ')' ? exp[1..^1] : exp;
+        }
+        //private string Trimb2(string exp)
```
xPaw, line 457: Remove this

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+        //{
+        //    return OffsetAtBranchExits.Count == 1 ? Trimb(exp) : exp;
+        //}
+
+
+        private readonly Dictionary<uint, string> ExternalVariablesPlaceholderNames = new();
+        private readonly Dictionary<uint, string> LocalVariableNames = new();
+
+        // naming external variables EXT, EXT2, EXT3,.. where not found
```
xPaw, line 466: Perhaps `Unknown` is better?

Frenzoid (approved):

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: I found them I'll write them in

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: ok

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: the array just converts an opcode to its string representation for the range 0xD to 0x17, hence indexes below 13 are not used  (it's only used on line 344 in the function ProcessOps)

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: you mean the comment .. right. I'll take it out

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: yea ok. This should hopefully never be seen if the vmat_c file is correct. I'll make the change

xPaw (approved):

ValveResourceFormat/Resource/ResourceTypes/Material.cs
```diff
@@ -79,6 +73,41 @@ public override void Read(BinaryReader reader, Resource resource)
             {
                 StringAttributes[kvp.GetProperty<string>("m_name")] = kvp.GetProperty<string>("m_pValue");
             }
+
+            // This is zero-length for all vmat files in Dota2 and HL archives
+            var textureAttributes = Data.GetArray<string>("m_textureAttributes");
+            if (textureAttributes.Length > 0) {
+                // Console.WriteLine("unexpected textureAttributes length");
```
Perryvw, line 80: Should probably be uncommented

ValveResourceFormat/Serialization/VfxEval.cs
Perryvw: Kinda ugly imo, why not just add 0xD to the index on line 344 then (with a comment explaining this stuff)?

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+            {
+                Expressions.Push($"{funcName}({Trimb(exp2)},{Trimb(exp1)})");
+                return;
+            }
+            var exp3 = Expressions.Pop();
+            if (nrArguments == 3)
+            {
+                // trim or not to trim ...
+                Expressions.Push($"{funcName}({Trimb(exp3)},{Trimb(exp2)},{Trimb(exp1)})");
+                // expressions.Push($"{funcName}({exp3},{exp2},{exp1})");
```
Perryvw, line 432: Commented line should be removed?

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -0,0 +1,503 @@
...
+                return;
+            }
+            var exp4 = Expressions.Pop();
+            if (nrArguments == 4)
+            {
+                Expressions.Push($"{funcName}({Trimb(exp4)},{Trimb(exp3)},{Trimb(exp2)},{Trimb(exp1)})");
+                return;
+            }
+
+            throw new Exception("this cannot happen!");
```
Perryvw, line 442:
```suggestion
            throw new Exception($"Dynamic expression: Unexpected number of arguments ({nrArguments}) for function ${funcName}");
```

ValveResourceFormat/Resource/ResourceTypes/Material.cs
robert-hoff: I wasn't sure what this parameter was so I scanned the dota2 and hl archives for it and found it was 0 in all cases. My guess is that there is a provision for it to do something but because of lack of data I couldn't say anything about what the effects a non-zero value would have. Guessing from the name a non-zero value might point to additional variables. Please confirm what you'd like to do with it

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: I took this out. I supplied some comments about the bracket removal instead. For example 1+2+3+4 will decompile as ((1+2)+3)+4. The parser doesn't make any attempt to check associativity or precedence, so removing the outer bracket only is the best it can do

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: I made this change, this is primarily a developer note because it really cannot happen, the value for nrArguments passed to ApplyFunction(..) is taken from the FUNCTION_REF lookup which only has values in the range [0,4]

xPaw: It cannot happen until it does. And when it does, the message should be reasonable :)

ValveResourceFormat/Resource/ResourceTypes/Material.cs
xPaw: Uncomment it, so if it happens we can at least see it and go from there.

robert-hoff: sounds good

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: yea agree, wrote it as a dictionary

PR #372, opened 2021-08-24
VfxEval update for parsing dynamic expressions from the shader code

robert-hoff:
Need to interpret the 1A and 1D opcodes to support parsing the shader code Gave these the names COND and EVAL. COND are usually seen as part of conditional statements and always returns a boolean. EVAL looks like evaluation of external functions, it returns a float usually, sometimes a vector

### DIFF
--- a/ValveResourceFormat/Serialization/VfxEval.cs
+++ b/ValveResourceFormat/Serialization/VfxEval.cs
@@ -119,32 +119,24 @@ namespace ValveResourceFormat.Serialization.VfxEval
         // check on each OPS if we are exiting a branch,
         // when we do we should combine expressions on the stack
         private readonly Stack<uint> OffsetAtBranchExits = new();
-
-        private readonly Dictionary<uint, string> ExternalVariablesPlaceholderNames = new();
         private readonly Dictionary<uint, string> LocalVariableNames = new();
 
         // build a dictionary of the external variables seen, passed as 'renderAttributesUsed'
         private static readonly ConcurrentDictionary<uint, string> ExternalVarsReference = new();
 
         // The 'return' keyword in the last line of a dynamic expression is optional (it is implied where absent)
-        // omitReturnStatement controls whether it is shown
-        private readonly bool omitReturnStatement;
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -438,13 +473,18 @@ private static string Trimb(string exp)
...
         // naming external variables UNKNOWN, UNKNOWN2, UNKNOWN3,.. where not found
+        // if showMurmurForUnknowns is enabled return them in the form UNKNOWN[e46d252d]
         private string GetExternalVarName(uint varId)
         {
             ExternalVarsReference.TryGetValue(varId, out var varKnownName);
             if (varKnownName != null)
             {
                 return varKnownName;
             }
+            if (showMurmurForUnknowns)
```
xPaw, line 484: Shouldn't this go inside the if to know that its actually unknown?

And is there any reason not to always return murmurs?

ValveResourceFormat/Serialization/VfxEval.cs
```diff
@@ -371,7 +400,13 @@ private void ProcessOps(OPCODE op, BinaryReader dataReader)
                 }
                 var finalExp = Expressions.Pop();
                 finalExp = Trimb(finalExp);
-                DynamicExpressionList.Add($"return {finalExp};");
+                if (omitReturnStatement)
+                {
+                    DynamicExpressionList.Add($"{finalExp}");
```
xPaw, line 405:
```suggestion
                    DynamicExpressionList.Add(finalExp);
```

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: Yea, always returning the murmur is good idea, will make this change

ValveResourceFormat/Serialization/VfxEval.cs
robert-hoff: got it

PR #373, opened 2021-08-28
Save and restore main window position and dimensions

robert-hoff:
Save and restore VRF window position and size between application runs

### DIFF
--- a/GUI/MainForm.cs
+++ b/GUI/MainForm.cs
@@ -34,15 +34,6 @@ namespace GUI
             LoadAssetTypes();
             InitializeComponent();
 
-            Settings.Load();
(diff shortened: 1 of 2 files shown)


xPaw:
Does this need to handle maximized state separately?

robert-hoff:
Yea we probably need to handle it separately - I'll look into it. I realised also it would be a good idea to do checks against the available screensize because if VRF was closed on a separate monitor it can happen that it becomes placed outside the available xy space

robert-hoff:
This should be a bit more general - I found that to get the window to restore to the last normal position (from a maximized state) the settings need to be applied after the form has been rendered, so I transferred the code into the OnShown() method

xPaw:
Merged in master

PR #406, opened 2022-03-18
Support for parsing vcs v62

robert-hoff:


ValveResourceFormat/CompiledShader/ShaderFile.cs
```diff
@@ -16,6 +16,7 @@ public class ShaderFile : IDisposable
         public const uint LZMA_DELIM = 0x414D5A4C;
         public const int ZSTD_COMPRESSION = 1;
         public const int LZMA_COMPRESSION = 2;
+        public const int UNCOMPRESSED = 3;
```
xPaw, line 19: This isn't the actual value in the shader file, is it? If so, uncompressed should be 0 instead, then you can check that compression>0

robert-hoff: set it to 0, like that ok?

PR #425, opened 2022-05-16
bug fixes

danggui1995:
1.fix vmdl export exception
2.fix shader RegexInclude not match

### DIFF
--- a/GUI/Types/Renderer/ShaderLoader.cs
+++ b/GUI/Types/Renderer/ShaderLoader.cs
@@ -15,7 +15,7 @@ namespace GUI.Types.Renderer
     {
         private const string ShaderDirectory = "GUI.Types.Renderer.Shaders.";
         private const int ShaderSeed = 0x13141516;
-        private static Regex RegexInclude = new Regex(@"^#include ""(?<IncludeName>[^""]+)""$", RegexOptions.Multiline);
+        private static Regex RegexInclude = new Regex(@"^#include ""(?<IncludeName>[^""]+)""", RegexOptions.Multiline);
         private static Regex RegexDefine = new Regex(@"^#define param_(?<ParamName>\S+) (?<DefaultValue>\S+)", RegexOptions.Multiline);
 
 #if !DEBUG_SHADERS || !DEBUG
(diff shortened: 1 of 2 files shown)


SinZ163:
A pull request is just merging branch A to branch B, you don't need to close and create new pull requests if you are trying to merge the exact same branch, it will update when you do more commits

danggui1995:
Got it, thanks.

GUI/Types/Renderer/ShaderLoader.cs
```diff
@@ -15,7 +15,7 @@ public class ShaderLoader
     {
         private const string ShaderDirectory = "GUI.Types.Renderer.Shaders.";
         private const int ShaderSeed = 0x13141516;
-        private static Regex RegexInclude = new Regex(@"^#include ""(?<IncludeName>[^""]+)""$", RegexOptions.Multiline);
+        private static Regex RegexInclude = new Regex(@"^#include ""(?<IncludeName>[^""]+)""", RegexOptions.Multiline);
         private static Regex RegexDefine = new Regex(@"^#define param_(?<ParamName>\S+) (?<DefaultValue>\S+)", RegexOptions.Multiline);
```
xPaw, line 19: Why?

GUI/Types/Renderer/ShaderLoader.cs
```diff
@@ -15,7 +15,7 @@ public class ShaderLoader
     {
         private const string ShaderDirectory = "GUI.Types.Renderer.Shaders.";
         private const int ShaderSeed = 0x13141516;
-        private static Regex RegexInclude = new Regex(@"^#include ""(?<IncludeName>[^""]+)""$", RegexOptions.Multiline);
```
xPaw, line 18: Why?

danggui1995: `GL.GetShader(vertexShader, ShaderParameter.CompileStatus, out var shaderStatus);`
Didn't you get the exception? I come to this exception and think the $ will not match when the next char is \r\n

xPaw: I added an optional `\r` there.

Can you test latest master for both that and the exporting code? I changed it to use same code as the renderer so it doesn't need to be updated in two places.

danggui1995: all tested ok

PR #426, opened 2022-06-01
support export model in folder

danggui1995:
Exporting folder will get model as vmdl file while I need gltf.


ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -2,14 +2,14 @@
...
-using ValveResourceFormat;
+using SteamDatabase.ValvePak;
 using ValveResourceFormat.ResourceTypes;
 
 namespace ValveResourceFormat.IO
 {
     public static class FileExtract
     {
-        public static Span<byte> Extract(Resource resource)
+        public static Span<byte> Extract(Resource resource, Package package = null, string targetPath = null)
```
xPaw, line 12: I think a better design would be to change this to ienumerable and yield more resources if needed. Also why do you need to pass a package, it's already set on the exporter object, is it not?

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -86,6 +117,14 @@ public static string GetExtension(Resource resource)
                 case ResourceType.PanoramaVectorGraphic: return "svg";
                 case ResourceType.Texture: return "png";
 
+                case ResourceType.Mesh:
+                case ResourceType.Model:
+                case ResourceType.World:
+                case ResourceType.WorldNode:
+                {
+                    return "gltf";
```
xPaw, line 125: Should probably use glb for the time being, gltf has some issues with huge files.

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -48,9 +48,40 @@ public static Span<byte> Extract(Resource resource)
...
+                                exporter.ExportToFile(resource.FileName, targetPath, new Mesh(resource));
+                                break;
+                            case ResourceType.Model:
+                                exporter.ExportToFile(resource.FileName, targetPath, (Model)resource.DataBlock);
+                                break;
+                            case ResourceType.WorldNode:
+                                exporter.ExportToFile(resource.FileName, targetPath, (WorldNode)resource.DataBlock);
+                                break;
+                            case ResourceType.World:
+                                exporter.ExportToFile(resource.FileName, targetPath, (World)resource.DataBlock);
```
xPaw, line 74: That being said, this doesn't actually work, does it? I don't see it being saved to a file.

xPaw:
Should figure out how to combine it with this code so there's no duplication. 
https://github.com/SteamDatabase/ValveResourceFormat/blob/c02530e3f96962d17918fbea2b7acdde4af82a03/GUI/Types/Exporter/ExportFile.cs#L69

xPaw:
Fixed by #448

PR #433, opened 2022-07-13
Correctly export material attributes

kristiker:
Getting back correct attributes now. Also fixed vec4 params so they all get parsed back correctly (compiler expects floats).
image

### DIFF
--- a/GUI/Types/Viewers/Resource.cs
+++ b/GUI/Types/Viewers/Resource.cs
@@ -2,7 +2,6 @@ using System;
 using System.ComponentModel;
 using System.Drawing;
 using System.IO;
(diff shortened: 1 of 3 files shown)


kristiker:
`SubrectDefinition` could be saved too but seems to be a bit more deeply linked:
In REDI:
```
ResourceEditStringData_t
{
	CResourceString m_Name = "SubrectDefinition"
	CResourceString m_String = "materials/architecture/bricks_hs.rect"
}
```
In Vmat:
```
ToolAttributes
{
	SubrectDefinition "materials/architecture/bricks_hs.rect"
}
```

PR #437, opened 2022-07-24
Adding separate image export option for spritesheets

Nibuja05:
- Adds the decompiler cli option `-t|--texture_split`
  - if option is given **and** output path is given **and** texture (.vtex_c) file is a spritesheet, it tries to export splitted images
  - exported images will have the same name with the following name signature: `NAME_seqXX_YY.png` (XX is sequence index; YY is frame index [if there are more than 1 frame per sequence])
- Adds `FileExtract.ExtractTextureSheet` function, that exports image data for all available sequences and frames

Known issues:
- Without changes to `DumpFile`, `OutputFile` is overwritten in order to save the files with different names.
- No integration for GUI yet; only works for Decompiler (although `ExtractTextureSheet` could probably be reused there as well)

### DIFF
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -13,6 +13,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using McMaster.Extensions.CommandLineUtils;
 using SteamDatabase.ValvePak;
+using SkiaSharp;
 using ValveResourceFormat;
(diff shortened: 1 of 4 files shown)


xPaw:
Implemented by #446

xPaw:
Duplicate of #446 (marking for github)

PR #438, opened 2022-07-24
Export model folders to glTF in GUI

kristiker:
* Clean up & finish #426


GUI/Forms/ExtractProgressForm.cs
```diff
@@ -144,10 +162,24 @@ private async Task ExtractFilesAsync()
...
+                    await using (stream.ConfigureAwait(false))
+                    {
+                        await stream.WriteAsync(output, cancellationTokenSource.Token).ConfigureAwait(false);
+                    }
+
+                    foreach(var child in extractedResource.Children)
+                    {
+                        var childPath = Path.Combine(outFolder, child.FileName);
+                        Directory.CreateDirectory(Path.GetDirectoryName(childPath));
+                        File.WriteAllBytes(childPath, child.Data);
```
kristiker, line 177: Needs async write?

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -1,16 +1,39 @@
...
+
+        public ExtractedResource()
+        {
+            Children = new List<ChildExtractedResource>();
+        }
+    }
+
+    public class ChildExtractedResource
+    {
+        public byte[] Data { get; set; }
```
xPaw, line 24: Aren't you going to consume a lot of memory trying to keep all the child resources in memory instead of writing them to disk as they come?

GUI/Forms/ExtractProgressForm.cs
xPaw: Yes.

ValveResourceFormat/IO/FileExtract.cs
xPaw: Also am I blind, where does ChildExtractedResource even get created?

kristiker: Example usage: https://github.com/kristiker/ValveResourceFormat/blob/f7d417652db94b5cc5a048a1f5da45a970d99e63/ValveResourceFormat/IO/FileExtract.cs#L93

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -41,6 +49,29 @@ public class GltfModelExporter
...
+                    ExportToFile(resource.FileName, targetPath, (VModel)resource.DataBlock);
+                    break;
+                case ResourceType.WorldNode:
+                    ExportToFile(resource.FileName, targetPath, (VWorldNode)resource.DataBlock);
+                    break;
+                case ResourceType.World:
+                    ExportToFile(resource.FileName, targetPath, (VWorld)resource.DataBlock);
+                    break;
+            }
+        }
```
xPaw, line 73: Add throw if not supported

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -29,6 +29,14 @@ public class GltfModelExporter
     {
         private const string GENERATOR = "VRF - https://vrf.steamdb.info/";
 
+        public static readonly ISet<ResourceType> ResourceTypesThatAreGltfExportable = new HashSet<ResourceType>()
```
xPaw, line 32: private

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -40,48 +68,52 @@ public static Span<byte> Extract(Resource resource)
...
-                            data = Encoding.UTF8.GetBytes(resource.DataBlock.ToString());
+                            extract.Data = Encoding.UTF8.GetBytes(resource.DataBlock.ToString());
                         }
 
                         break;
                     }
             }
 
-            return data;
+#if DEBUG_CONTENTSUBFILE
```
xPaw, line 112: Remove this

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -34,6 +38,11 @@ public ExtractProgressForm(Package package, TreeNode root, string path, bool dec
             this.root = root;
             this.path = path;
             this.decompile = decompile;
+
+            gltfExporter = new GltfModelExporter()
+            {
+                FileLoader = new BasicVpkFileLoader(package)
```
xPaw, line 44: Can't `AdvancedGuiFileLoader` be used in GUI? I think it should be pass in the `VrfGuiContext`.

We have `ExportData` class already, perhaps it can be reused.

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -22,6 +22,10 @@ public partial class ExtractProgressForm : Form
         private CancellationTokenSource cancellationTokenSource;
         private int initialFileCount;
 
+        private ContentFile contentFile;
```
xPaw, line 25: This can be local?

Decompiler/Decompiler.cs
```diff
@@ -412,11 +412,13 @@ private void ProcessFile(string path, Stream stream, string originalPath = null)
 
                 if (OutputFile != null)
                 {
-                    var data = FileExtract.Extract(resource);
+                    var contentFile = FileExtract.Extract(resource);
 
                     var filePath = Path.ChangeExtension(path, extension);
 
-                    DumpFile(filePath, data);
+                    // TODO: fix exporting subfiles to a non-folder --output
```
xPaw, line 419: How hard would it be to fix? What currently happens?

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -144,10 +162,30 @@ private async Task ExtractFilesAsync()
...
+                    Directory.CreateDirectory(Path.GetDirectoryName(subFilePath));
+
+                    byte[] subFileData;
+                    try
+                    {
+                        subFileData = contentSubFile.Extract();
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine($"Failed to extract subfile '{contentSubFile.FileName}' - {e.Message}");
```
xPaw, line 183: Write to `Console.Error`

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -1,16 +1,44 @@
...
 using System.IO;
 using System.Text;
 using SkiaSharp;
 using ValveResourceFormat.ResourceTypes;
 
 namespace ValveResourceFormat.IO
 {
+    public class ContentFile
+    {
+        public byte[] Data { get; set; }
```
xPaw, line 12: > I had to undo the Span<byte> change as it wouldn't play nice with the async function that exports whole folders.

What's the problem?

xPaw (changes requested):

ValveResourceFormat/IO/FileExtract.cs
kristiker: I tried to make ContentFile a ref struct (to hold Span) but it tells me that is stack based and cant be used in async functions.

GUI/Forms/ExtractProgressForm.cs
kristiker: Not sure. The rest of the function only deals with the current package.

This all needs to be moved to `GUI/Types/Exporter/ExportFolder.cs` and then something like `ExportData` could be used to clarify if the folder needs to be GltfExported or just Extract.

Currently `ExtractProgressForm.ExtractFilesAsync` exports supported files to gltf and extracts the rest.

Decompiler/Decompiler.cs
kristiker: > What currently happens?

Currently any *possible* subfiles are skipped. Vpk dumping & folder dumping works well but there is an issue when doing 

`decompiler -i "sprite.vtex_c" -o "sprite.png" -d`

DumpFile has some busted `filePath` logic and sends the last subfile to `sprite.png` as well, instead of the subfile path that I'm giving to it. I just can't fix it in a nice way.

xPaw: I wouldn't say it's busted given the restrictions we had before. I'd say one way of fixing it would be to dump the 'main' file to the given output path as-is, and for subfiles either name them similarly to the output path (but append number to it or whatever), or force useOutputAsDirectory to true.

kristiker: I would like to deal with this after we've got some subfiles rolling. Decompiler is very folder/pak oriented and it's hard not to break those uses in the process.

xPaw: If I were to merge this pr as-is, will it break current scripts?

How do we avoid the breakage?

kristiker: Yes, works as before and vpk and folder decompiling will create extra files where available.

> How do we avoid the breakage?

Imo it shouldn't be allowed to put a file as -o. The code can be much cleaner then.

kristiker: Nevermind. Did a slight refactor to the original and fixed it.

xPaw:
Can you update the pr description please. Do I understand correctly that this PR now adds support for decompiling folders from gui?

kristiker:
Yes. Will probably move it to a new pr once it's done.

PR #443, opened 2022-09-18
Basic vsnd_c phoneme reading

kristiker:
Related to this https://developer.valvesoftware.com/wiki/Phoneme_Tool/data_format

Only reads the WORDS section, I couldn't figure out how other stuff is packed.

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Sound.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Sound.cs
@@ -10,7 +10,7 @@ namespace ValveResourceFormat.ResourceTypes
     public struct EmphasisSample
     {
         public float Time { get; }
-        public float Value { get;}
+        public float Value { get; }
     }
 
     public struct PhonemeTag
@@ -94,7 +94,7 @@ namespace ValveResourceFormat.ResourceTypes
 
         public float Duration { get; private set; }
 
-        public Sentence Sentence {get; private set; }
+        public Sentence Sentence { get; private set; }
 
         public uint StreamingDataSize { get; private set; }
 
@@ -170,10 +170,10 @@ namespace ValveResourceFormat.ResourceTypes
             Duration = reader.ReadSingle();
 
             var sentenceOffset = reader.ReadInt64();
-            
+
             if (sentenceOffset != 0)
                 sentenceOffset = reader.BaseStream.Position + sentenceOffset;
-            
+
             // Skipping over m_pHeader
             reader.BaseStream.Position += 4;
 
@@ -222,10 +222,12 @@ namespace ValveResourceFormat.ResourceTypes
         private void ReadPhonemeStream(BinaryReader reader, long sentenceOffset)
         {
             if (sentenceOffset == 0)
+            {
                 return;
-            
+            }
+
             Reader.BaseStream.Position = sentenceOffset;
-            
+
             var numPhonemeTags = reader.ReadInt32();
 
             var a = reader.ReadInt32(); // numEmphasisSamples ?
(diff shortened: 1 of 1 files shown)


ValveResourceFormat/Resource/ResourceTypes/Sound.cs
```diff
@@ -144,13 +169,19 @@ public override void Read(BinaryReader reader, Resource resource)
             SampleCount = reader.ReadUInt32();
             Duration = reader.ReadSingle();
 
-            // Skipping over m_Sentence (CSentence_t) and m_pHeader
-            reader.BaseStream.Position += 12;
+            var sentenceOffset = reader.ReadInt64();
```
xPaw, line 172: Is it actually long?

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
```diff
@@ -144,13 +169,19 @@ public override void Read(BinaryReader reader, Resource resource)
...
+            
+            if (sentenceOffset != 0)
```
xPaw, line 174: Add brackets, and run format document in VS

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
```diff
@@ -144,13 +169,19 @@ public override void Read(BinaryReader reader, Resource resource)
...
+                sentenceOffset = reader.BaseStream.Position + sentenceOffset;
+            
+            // Skipping over m_pHeader
+            reader.BaseStream.Position += 4;
 
             StreamingDataSize = reader.ReadUInt32();
 
             if (resource.Version < 1)
```
xPaw, line 182: Can you flip the ifs so there's no early returns?

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
```diff
@@ -179,12 +211,51 @@ public override void Read(BinaryReader reader, Resource resource)
...
-            LoopEnd = reader.ReadInt32();
+            Sentence = new Sentence();
+
+            Sentence.RunTimePhonemes = new PhonemeTag[numPhonemeTags];
+
+            for (var i = 0; i < numPhonemeTags; i++)
+            {
+                var startTime = reader.ReadSingle();
+                var endTime = reader.ReadSingle();
+                var phonemeCode = (UInt16)reader.ReadInt32();
```
xPaw, line 248: Why?

kristiker: Valve aligns it to 4 bytes. Should I really ReadUInt16 and skip 2?

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
kristiker: Yeah probably isn't. Will revert it now. Reads fine though on all vsnds (next int is always 0x00).

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
xPaw: If it's alignment bytes, skip them separately.

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
xPaw: There are a bunch of zeroes, like all the others where we check for zero and throw if it's not (to find any).

ValveResourceFormat/Resource/ResourceTypes/Sound.cs
kristiker: Done.

PR #446, opened 2022-10-11
Export individual texture sequences based on `SpritesheetData`

kristiker:
Better implements #437

There is also a simplistic [MakeSheet file](https://developer.valvesoftware.com/wiki/Animated_Particles#Creating_an_MKS_File) that is exported.

### DIFF
--- a/ValveResourceFormat/IO/FileExtract.cs
+++ b/ValveResourceFormat/IO/FileExtract.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Text;
 using SkiaSharp;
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -63,11 +64,67 @@ public static ContentFile Extract(Resource resource)
 
                 case ResourceType.Texture:
                     {
-                        using var bitmap = ((Texture)resource.DataBlock).GenerateBitmap();
+                        var bitmap = ((Texture)resource.DataBlock).GenerateBitmap();
```
xPaw, line 67: So we need to keep it around for ImageExtract? That sounds like it never gets explicitly exposed.

kristiker: Yup it's left up to the GC.

xPaw:
There are sprites where multiple frames can use the same images (materials/particle/fire/fire_animated.vtex_c), should we deduplicate images by their coordinates?

PR #448, opened 2022-10-13
Models use gltf exporter when doing a folder export

kristiker:
Models will export to GLB instead of VMDL when doing a folder export with the GUI.
image

Also merged the two options into one since you can pick via the Save File dialog anyway.
image


Decompiler/Decompiler.cs
```diff
@@ -865,29 +865,21 @@ private void DumpVPK(string parentPath, Package package, string type, Dictionary
 
                         extension = FileExtract.GetExtension(resource) ?? type[..^2];
 
-                        // TODO: Hook this up in FileExtract
-                        if (resource.ResourceType == ResourceType.Mesh || resource.ResourceType == ResourceType.Model)
+                        if (GltfModelExporter.CanExport(resource))
```
kristiker, line 868: Missed by #344 or is it more helpful that Decompiler only does models?

xPaw: Just missed.

PR #449, opened 2022-10-15
Further improve texture sheet exporting

kristiker:
- [x] Only export unique frames.
- [x] Write correct mks for sheets with `packmode rgb+a`.
- [x] Display the mks on a GUI tab.
- [x] Sprite textures (non-child) write .vtex as their main content file, also display on a GUI tab.
    - [x] Child textures such as material maps export the png only (behaviur pre this PR).
 - [x] Free bitmap off memory when exporting sheets too.

### DIFF
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -378,7 +378,7 @@ namespace Decompiler
                     if (OutputFile == null)
                     {
                         // Test extraction code flow while collecting stats
-                        FileExtract.Extract(resource);
+                        using FileExtract.Extract(resource);
                     }
 
                     if (!string.IsNullOrEmpty(info))
@@ -849,7 +849,7 @@ namespace Decompiler
                 Console.WriteLine("\t[archive index: {0:D3}] {1}", file.ArchiveIndex, filePath);
 
                 package.ReadEntry(file, out var output);
-                var contentFile = default(ContentFile);
+                ContentFile contentFile;
 
                 if (type.EndsWith("_c", StringComparison.Ordinal) && Decompile)
                 {
@@ -926,9 +926,11 @@ namespace Decompiler
             {
                 foreach (var contentSubFile in contentFile.SubFiles)
                 {
-                    DumpFile(Path.Combine(Path.GetDirectoryName(path), contentSubFile.FileName), contentSubFile.Extract());
+                    DumpFile(Path.Combine(Path.GetDirectoryName(path), contentSubFile.FileName), contentSubFile.Extract.Invoke());
                 }
             }
+
+            contentFile.Dispose();
         }
 
         private static void DumpFile(string path, ReadOnlySpan<byte> data)
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -378,7 +378,7 @@ namespace Decompiler
                     if (OutputFile == null)
                     {
                         // Test extraction code flow while collecting stats
-                        using FileExtract.Extract(resource);
+                        FileExtract.Extract(resource).Dispose();
                     }
 
                     if (!string.IsNullOrEmpty(info))
@@ -849,7 +849,7 @@ namespace Decompiler
                 Console.WriteLine("\t[archive index: {0:D3}] {1}", file.ArchiveIndex, filePath);
 
                 package.ReadEntry(file, out var output);
-                ContentFile contentFile;
+                ContentFile contentFile = null;
 
                 if (type.EndsWith("_c", StringComparison.Ordinal) && Decompile)
                 {
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -378,7 +378,7 @@ namespace Decompiler
                     if (OutputFile == null)
                     {
                         // Test extraction code flow while collecting stats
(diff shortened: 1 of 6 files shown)


ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -64,67 +64,14 @@ public static ContentFile Extract(Resource resource)
...
-                        contentFile.Data = png.ToArray();
-
-                        var spriteSheetData = ((Texture)resource.DataBlock).GetSpriteSheetData();
-
-                        if (spriteSheetData is null)
+                        var textureExtract = new TextureExtract(resource.FileName, (Texture)resource.DataBlock);
+                        if (IsChildResource(resource))
                         {
-                            bitmap.Dispose();
+                            contentFile.Data = textureExtract.ToPngImage();
```
kristiker, line 70: Rip dispose. If sheets aren't doing it, does it matter?

xPaw: The problem here is that skia has unmanaged objects, so solely relying on GC is not ideal.

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -182,6 +129,17 @@ public static ContentFile Extract(Resource resource)
...
 
+        public static bool IsChildResource(Resource resource)
+        {
+            if (resource.EditInfo is ResourceEditInfo2 redi2)
+            {
+                return redi2.SearchableUserData.GetProperty<long>("IsChildResource") == 1;
+            }
+
+            var extraIntData = (Blocks.ResourceEditInfoStructs.ExtraIntData)resource.EditInfo.Structs[ResourceEditInfo.REDIStruct.ExtraIntData];
+            return extraIntData.List.Where(x => x.Name == "IsChildResource").Select(x => x.Value).FirstOrDefault() == 1;
```
xPaw, line 140:
```suggestion
            return extraIntData.List.FirstOrDefault(x => x.Name == "IsChildResource")?.Value == 1;
```

I didn't test it, but you should be able to get something like this to work.

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -191,7 +149,14 @@ public static string GetExtension(Resource resource)
                 case ResourceType.PanoramaTypescript: return "js";
                 case ResourceType.PanoramaStyle: return "css";
                 case ResourceType.PanoramaVectorGraphic: return "svg";
-                case ResourceType.Texture: return "png";
+
+                case ResourceType.Texture:
+                    if (IsChildResource(resource))
```
xPaw, line 154: And if it's not? Why is the check necessary?

GUI/Types/Viewers/Resource.cs
```diff
@@ -394,46 +394,56 @@ public TabPage Create(VrfGuiContext vrfGuiContext, byte[] input)
...
+                        break;
+
+                    case ResourceType.Texture:
+                        {
+                            if (FileExtract.IsChildResource(resource))
+                            {
+                                break;
+                            }
+
+                            var textureExtract = new TextureExtract(resource.FileName, (Texture)resource.DataBlock);
```
xPaw, line 437: this will call GenerateBitmap, which is slow and you don't need here.

ValveResourceFormat/IO/FileExtract.cs
kristiker: Fixed! Disposing everything now.

GUI/Types/Viewers/Resource.cs
kristiker: Fixed, bitmap is only generated when forming the content file for export.

Decompiler/Decompiler.cs
```diff
@@ -378,7 +378,7 @@ private void ProcessFile(string path, Stream stream, string originalPath = null)
                     if (OutputFile == null)
                     {
                         // Test extraction code flow while collecting stats
-                        using FileExtract.Extract(resource);
+                        FileExtract.Extract(resource).Dispose();
```
xPaw, line 381: `using var _ =  FileExtract.Extract(resource);` is how you do it.

Your dispose won't be called if extract throws.

Decompiler/Decompiler.cs
```diff
@@ -889,6 +889,7 @@ private void DumpVPK(string parentPath, Package package, string type, Dictionary
                     catch (Exception e)
                     {
                         LogException(e, filePath, package.FileName);
+                        contentFile?.Dispose();
```
xPaw, line 892: This seems suspicious because there's using contentFile below, should contentFile be reset to null, outputfile set to null, or check that contentFile isn't null below?

kristiker: Oh it still goes through... lol. I assume it was meant to output the file "as is" in case contentFile extraction fails. In that case a null check should do.

xPaw: Generally for disposables it should be wrapped in using for the all the code that accesses the variable.

That or, you can manually write out try/finally with a dispose call. Right now this code still looks like it could miss a dispose if something throws in the middle.

kristiker: Should be fine now. Realistically, if those things throw the program would quit anyway though, I think.

PR #451, opened 2022-10-27
Unpack textures when exporting vmat_c

kristiker:
From asset without content:

To recompiled and working:

- [x] Cleanup.
- [ ] Add more shader mappings.
- [ ] Alternatively, gather packing data from actual shader files.
- [ ] Fix any bad color transformations (some BC7 stuff, alpha premul). `gman_pants.vmat`

Folder extracts:
- [x] Extract reused textures only once.
- [x] Extract materials first, then textures, skipping any textures that were first handled by MaterialExtract.
- [x] Textures sitting on a whole different folder get written to that folder, unlike single file extract which places all textures next to the material. This makes links inside materials work 100%. GUI only.

### DIFF
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -811,9 +811,18 @@ namespace Decompiler
 
         private void DumpVPK(string parentPath, Package package, string type, Dictionary<string, uint> manifestData)
         {
-            if (ExtFilterList != null && !ExtFilterList.Contains(type))
+            var allowSubFilesFromExternalRefs = true;
+            if (ExtFilterList != null)
             {
-                return;
+                if (!ExtFilterList.Contains(type))
+                {
+                    return;
+                }
+
+                if (type == "vmat_c" && ExtFilterList.Contains("vmat_c") && !ExtFilterList.Contains("vtex_c"))
+                {
+                    allowSubFilesFromExternalRefs = false;
+                }
             }
 
             if (!package.Entries.ContainsKey(type))
@@ -912,7 +921,7 @@ namespace Decompiler
 
                         if (Decompile && contentFile is not null)
                         {
-                            DumpContentFile(filePath, contentFile);
+                            DumpContentFile(filePath, contentFile, allowSubFilesFromExternalRefs);
                         }
                         else
                         {
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -378,7 +378,7 @@ namespace Decompiler
                     if (OutputFile == null)
                     {
                         // Test extraction code flow while collecting stats
-                        using var _ = FileExtract.Extract(resource);
+                        using var _ = FileExtract.Extract(resource, null);
                     }
 
                     if (!string.IsNullOrEmpty(info))
@@ -420,7 +420,7 @@ namespace Decompiler
 
                 if (OutputFile != null)
                 {
-                    using var contentFile = FileExtract.Extract(resource);
+                    using var contentFile = FileExtract.Extract(resource, null);
 
                     path = Path.ChangeExtension(path, extension);
                     var outFilePath = GetOutputPath(path);
@@ -884,7 +884,7 @@ namespace Decompiler
                             continue;
                         }
 
-                        contentFile = FileExtract.Extract(resource);
+                        contentFile = FileExtract.Extract(resource, fileLoader);
                     }
                     catch (Exception e)
                     {
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -75,6 +75,7 @@ namespace GUI.Forms
                         extractProgressBar.Style = ProgressBarStyle.Marquee;
                     }));
 
+                    Console.WriteLine($"Folder export started to \"{path}\"");
                     CalculateFilesToExtract(root);
 
                     Invoke((Action)(() =>
@@ -208,7 +209,7 @@ namespace GUI.Forms
                     {
                         if (contentFile.Data.Length > 0)
                         {
-                            Console.WriteLine($"Writing content file: {outFilePath}");
+                            Console.WriteLine($"+ {outFilePath.Remove(0, path.Length + 1)}");
                             await File.WriteAllBytesAsync(outFilePath, contentFile.Data, cancellationTokenSource.Token).ConfigureAwait(false);
                         }
 
@@ -217,6 +218,7 @@ namespace GUI.Forms
                         {
                             foreach (var (refFileName, refContentFile) in contentFile.ExternalRefsHandled)
                             {
+                                Invoke(() => extractStatusLabel.Text = $"Extracting {refFileName}");
                                 extractedFiles.Add(refFileName);
                                 await ExtractSubfiles(Path.GetDirectoryName(refFileName), refContentFile).ConfigureAwait(false);
                             }
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -19,10 +19,18 @@ namespace GUI.Forms
         private readonly TreeNode root;
         private readonly string path;
         private readonly ExportData exportData;
+        private readonly Dictionary<string, Queue<PackageEntry>> filesToExtractSorted;
         private readonly Queue<PackageEntry> filesToExtract;
+        private readonly HashSet<string> extractedFiles;
         private readonly GltfModelExporter gltfExporter;
         private CancellationTokenSource cancellationTokenSource;
-        private int initialFileCount;
+
+        private static readonly List<ResourceType> ExtractOrder = new()
+        {
+            // Materials before textures
+            ResourceType.Material,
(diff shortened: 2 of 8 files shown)


ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,405 @@
...
+                vmat.SubFiles.Add(imageSubFile);
+            }
+        }
+
+        return vmat;
+    }
+
+    public static string OutTextureName(string texturePath, string desiredSuffix = null)
+    {
+        texturePath = Path.ChangeExtension(texturePath, null); // strip .vtex_c
```
xPaw, line 160: Verify that you're stripping the right things?

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,405 @@
...
+                TextureType = guessedTextureType,
+                TextureFileName = OutTextureName(texturePath),
+                Channels = "RGBA"
+            };
+        }
+    }
+
+    public static bool TryFigureOutNonStaticMap(string shader, string textureType, Dictionary<string, long> intParams, out string newTextureType)
+    {
+        if (shader == "vr_simple" && textureType == "g_tColor")
```
xPaw, line 247: Writing shader specific code is oof. Any reason this data doesn't fit in the mapping above?

ValveResourceFormat/IO/MaterialExtract.cs
kristiker: It should never fail with the current CompileMaterial implementation.

xPaw: *should*, but this will cause silent issues if it does change (or is somehow a given a different file).

Decompiler/Decompiler.cs
```diff
@@ -420,7 +420,7 @@ private void ProcessFile(string path, Stream stream, string originalPath = null)
 
                 if (OutputFile != null)
                 {
-                    using var contentFile = FileExtract.Extract(resource);
+                    using var contentFile = FileExtract.Extract(resource, null);
```
xPaw, line 423: perhaps a future improvement (not in this pr): get a file loader in here (vpk or from disk when parsing raw file).

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+{
+    R,
+    G,
+    B,
+    A,
+    _OneChannel,
+    GA,
+    _TwoChannels,
+    RGB,
+    All
```
xPaw, line 32: Call it RGBA?

It also feels like it should be possible to make this enum flags, and get rid of _OneChannel/_TwoChannels

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+            }
+
+            // Keep the hash to avoid conflicts
+            texturePath += "_" + textureParts[^1] + desiredSuffix;
+        }
+
+        return Path.ChangeExtension(texturePath, keepOriginalExtension ? textureParts[^2] : "png");
+    }
+
+    public IEnumerable<UnpackInfo> GetInputImagesForTexture(string textureType, string texturePath, bool ommitDefault,
```
xPaw, line 246:
```suggestion
    public IEnumerable<UnpackInfo> GetInputImagesForTexture(string textureType, string texturePath, bool omitDefault,
```

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+            Dictionary<string, long> intParams, Dictionary<string, Vector4> vectorParams)
+    {
+        var shader = material.ShaderName[..^4];
```
xPaw, line 249: Can you at least add comment what is being dropped here.

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+using System.Text;
+using ValveKeyValue;
+using ValveResourceFormat.Blocks;
+using ValveResourceFormat.ResourceTypes;
+
+namespace ValveResourceFormat.IO;
+
+using ChannelMapping = ValueTuple<Channel, string>;
+
+public struct UnpackInfo
```
xPaw, line 15: `MaterialUnpackInfo`

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+{
+    public string TextureType { get; init; }
+    public string FileName { get; init; }
+    public Channel Channel { get; init; }
+}
+
+public enum Channel
```
xPaw, line 22: `ColorChannel` or something? Channel is a bit generic in IO namespace. Or put these two inside of MaterialExtract.

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+
+        return Path.ChangeExtension(texturePath, keepOriginalExtension ? textureParts[^2] : "png");
+    }
+
+    public IEnumerable<UnpackInfo> GetInputImagesForTexture(string textureType, string texturePath, bool ommitDefault,
+            Dictionary<string, long> intParams, Dictionary<string, Vector4> vectorParams)
+    {
+        var shader = material.ShaderName[..^4];
+
+        if (TextureMappings.ContainsKey(shader) && TextureMappings[shader] is IDictionary<string, ChannelMapping[]> mappings && mappings.ContainsKey(textureType))
```
xPaw, line 251: Use TryGetValue, and why the `is` check? When can it possibly be not that.

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    public static string OutTextureName(string texturePath, bool keepOriginalExtension, string desiredSuffix = null)
+    {
+        if (!texturePath.EndsWith(".vtex", StringComparison.OrdinalIgnoreCase))
+        {
+            throw new ArgumentException($"{texturePath} must have .vtex", nameof(texturePath));
+        }
+        texturePath = Path.ChangeExtension(texturePath, null); // strip .vtex
+        texturePath = Path.ChangeExtension(texturePath, null); // strip .generated (if present)
```
xPaw, line 209: `if (texturePath.EndsWith(".generated", StringComparison.OrdinalIgnoreCase))`

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -0,0 +1,482 @@
...
+            attributes.Add(new KVObject(key, $"[{value.X:N6} {value.Y:N6} {value.Z:N6} {value.W:N6}]"));
+        }
+
+        foreach (var (key, value) in material.StringAttributes)
+        {
+            attributes.Add(new KVObject(key, value ?? string.Empty));
+        }
+
+        var attributesThatAreSystemAttributes = new HashSet<string>
+            {
```
xPaw, line 433: extra indent

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -60,15 +75,27 @@ protected override void OnShown(EventArgs e)
...
 
                     Invoke((Action)(() =>
                     {
                         extractProgressBar.Style = ProgressBarStyle.Continuous;
                     }));
 
-                    await ExtractFilesAsync().ConfigureAwait(false);
+                    if (Decompile)
+                    {
+                        foreach (var resourceType in ExtractOrder)
```
xPaw, line 88: Loop filesToExtractSorted directly?

Decompiler/Decompiler.cs
kristiker: Indeed, can't properly decompile a vmat_c from disk the way it is right now, as it won't find the textures.

ValveResourceFormat/IO/MaterialExtract.cs
kristiker: Even if `Path.ChangeExtension` does the same check and will leave it unchanged?

xPaw: ChangeExtension will strip anything if the file name has a dot in it.

ValveResourceFormat/IO/MaterialExtract.cs
kristiker: Putted them both inside MaterialExtract. I need Channel to be short

ValveResourceFormat/IO/MaterialExtract.cs
kristiker: Fixed!

GUI/Forms/ExtractProgressForm.cs
kristiker: Thanks, I wouldn't have spotted this ever.

xPaw: Well crap, I forgot about that.

Issue #287, opened 2020-09-01
Switch to .NET native Half type


xPaw:
https://devblogs.microsoft.com/dotnet/introducing-the-half-type/

This should remove `ValveResourceFormat.HalfTypeHelper`

ch-asimakopoulos:
Can I take a crack at this?

xPaw:
Sure. Keep in mind .NET 5 does not have extension methods to read halfs in BinaryReader.

ch-asimakopoulos:
@xPaw  how about this?

xPaw:
You can give it a try where half type helper is called, I am unsure.

PR #463, opened 2022-12-06
Fix CCompressedDeltaVector3 and CCompressedStaticVector3 reading

moofMonkey:
This PR fixes CCompressedStaticVector3 and CCompressedDeltaVector3 reading.

All compressed data that is marked static is meant to only have one frame, so it should return 0 from AnimDecoder#Size to avoid seeking to frames past 0 - as it was done with CCompressedStaticFullVector3 and others.
CCompressedDeltaVector3 is basically CCompressedStaticFullVector3 combined with CCompressedAnimVector3.

Example for CCompressedDeltaVector3: Dota 2 `models/creeps/lane_creeps/creep_bad_ranged/lane_dire_ranged.vmdl` attack2 and attack3 animations.
I couldn't find an example for CCompressedStaticVector3 right away, but it is easy to figure out by comparing CCompressedStaticVector3 and CCompressedAnimVector3 in debug bins (specifically `libanimationsystem.dylib`), as they only differ at the start by seeking to the current frame.

Visual differences, taken from sandbox.babylonjs.com with resulting glTF with default camera position and angles at frame 30.04:
Mostly it's just skeleton being off.

attack2:
Current implementation:
attack2_old
This PR:
attack2_new

attack3:
Current implementation:
attack3_old
This PR:
attack3_new


xPaw:
Huh it was this simple? Is it always actually one frame, and not possibly deltas against other frames like I assumed?

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
```diff
@@ -270,12 +273,22 @@ private void ReadSegment(long frame, IKeyValueCollection segment, IKeyValueColle
...
                         outFrame.SetAttribute(boneNames[bone], channelAttribute, new Vector3(
                             containerReader.ReadSingle(),
                             containerReader.ReadSingle(),
                             containerReader.ReadSingle()));
                         break;
+                    case AnimDecoderType.CCompressedDeltaVector3:
+                        // Structure is | Bone 0 - Base Frame | Bone 1 - Base Frame | Bone 0 - Frame 0 | Bone 1 - Frame 0|
+                        if (containerReader.BaseStream.Position + (12 * numBones) < containerReader.BaseStream.Length)
+                        {
+                            containerReader.BaseStream.Position += 12 * numBones;
```
xPaw, line 285: So if I'm reading this right, it has a Vector3<float> for each bone stored (frame 0), and then the delta frames are Vector3<half>? Even the first frame has a delta of its own?

The only reason we really have a `containerReader.BaseStream.Length` check is because we don't fully understand decoders, so it doesn't overflow. But if this is correct implementation, this check is not needed here.

And you're sure the delta is always against the first frame and not the previous one? I'm not entirely sure that having a second binary reader (baseFrameContainerReader) is an elegant way of doing this.

xPaw: Whatever changes are required here to make it work correctly is for the best (I certainly think that swapping the switch/for is a good idea, probably can move each decoder into its own method).

The anim decoder is in dire need of proper implementation, it just about barely works currently. If you're willing to make it nicer here, I'm all for it.

moofMonkey:
> If you want, I could find various animation files with different decoder types for you to test (would need to hook something up in the cli's --stats scanner).

I already tested it in-game by checking it against `ent_attachments` and `ent_skeleton`, except with a different implementation - which is why there was a bug with incrementing position per every bone instead of once initially.
I'm not 100% sure, but from what I saw in-game and in the IDA - decoder implementations should be correct now.

xPaw:
Here's a few models that use CCompressedStaticVector3:

Parent file: dota 2 beta\game\core\pak01_dir.vpk
models/creatures/box_creature/box_creature_model.vmdl_c
I believe ref_pose uses it and it looks correct.


dota 2 beta\game\dota_addons\aghanim\models\heroes\aghanim\aghanim_model.vmdl_c
dota 2 beta\game\dota_addons\diretide2020\models\diretide_intro_layout.vmdl_c
dota 2 beta\game\dota_addons\diretide2020\models\diretide_intro_sky_dome.vmdl_c

moofMonkey:
I committed a rewrite of animations parser. Now it consumes less memory and takes less time to load the model, as it defers frame decoding until it's needed. Framerate is not impacted as far as I can tell.

Dota 2 `models/heroes/monkey_king/monkey_king.vmdl_c` is a great example of how that works, especially its `mk_showcase_intro` animation which is ~30 seconds long.

Behavior for Release under debugger for the following model and animation:
Old:
462MB before opening model, 982MB right after opening it with visible lag due to GC nor frame decoding, 990MB after selecting animation, 927MB after waiting a bit for GC with animation playing.
New:
462MB before opening model, 1000MB right after opening it, 1008MB after selecting animation, 815MB after waiting a bit for GC with animation playing.
Difference for Debug build (where GC is presumably different) is that old behavior uses 1558MB RAM at peak, while new one is about the same as Release (~1008MB).
So basically about the same peak RAM usage in Release, but less overall RAM usage. In Debug difference is much more noticeable.

I'm not sure whether I need to update the PR or make another one.

I'm not sure about the code style (compared to the rest of the repo, but I followed .editorconfig) and removed comments, and whether something could be done better in C#.
For example segment decoders would benefit from using ArraySegment, but I don't think it's worth additional code because of streams.

I'm also not sure about whether to use mutex (as it is using now), or to avoid reusing Frame objects (which will increase memory pressure).
One or the other is required in case animation gets played simultaneously with glTF export happening on the other thread due to the race condition.

Additional testing might be required, but exported glTF didn't change from previous commit for Dota 2 `models/creeps/lane_creeps/creep_bad_ranged/lane_dire_ranged.vmdl`.

moofMonkey:
Rebased to https://github.com/SteamDatabase/ValveResourceFormat/commit/09787d0e6ebc892cba7434f5e0995bc9487da0a2

xPaw:
I'd put all the classes into their own files, as we already have a separate folder and namespace for animations anyway.

I'm not convinced a mutex is actually needed here. How well does it work without a frame cache/mutex? I'd rather change the GUI so that it doesn't call GetFrame on the same frame.

moofMonkey:
> I'd put all the classes into their own files, as we already have a separate folder and namespace for animations anyway.

Sure, done.

> I'm not convinced a mutex is actually needed here. How well does it work without a frame cache/mutex? I'd rather change the GUI so that it doesn't call GetFrame on the same frame.

The problem is that GUI is not calling the `GetFrame` by itself, it's called via `GetAnimationMatricesAsArray` => `GetAnimationMatrices` => `GetTransformsAtTime`.
I changed behavior from using mutex to separating `DecodeFrame` and `GetFrame`, and making `GetFrame` private, so that glTF export won't interfere with the GUI - and will be a bit faster.
Everything should work well as long as `GetAnimationMatrices`/`GetAnimationMatricesAsArray` are only called from one thread at a time.

> Since the original data is a byte array, you could wrap it into Span and use methods like BitConverter.ToSingle. This would also remove the Dispose method, which currently is not implemented using IDisposable.

I forgot about leaving Dispose, thanks.
Spans didn't really work out because they can't be properly stored in classes nor used in lambdas (for LINQ), but I rewrote everything using ArraySegments and LINQ, performance/RAM usage didn't seem to change much, however GC pressure could be a bit lower.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/SegmentDecoders/CCompressedStaticVector3.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public CCompressedStaticVector3(ArraySegment<byte> data, int[] elements, AnimationDataChannel localChannel) : base(elements, localChannel)
+        {
+            // Static data has only one frame of data, so prefetch all data to avoid unnecessary GC
+            Data = Enumerable.Range(0, data.Count / (3 * 2))
+                .Select(i =>
+                {
+                    var offset = i * (3 * 2);
+                    return new Vector3(
+                        (float)BitConverter.ToHalf(data.Slice(offset + (0 * 2))),
```
xPaw, line 19: Did you test that BitConverter.ToHalf produces same results as HalfTypeHelper.Convert?

xPaw:
Most of the changes look all right, however I'm still unsure about the frame cache, especially since it seems to be not thread safe.

Perhaps this PR could be simplified to decode all the frames like it was before? That makes it trivial to merge most of the rewrite.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/SegmentDecoders/CCompressedStaticVector3.cs
moofMonkey: I tested it like that
```cs
DeltaData = Enumerable.Range(0, deltaData.Count / 2)
    .Select(i =>
    {
        var a = (float)BitConverter.ToHalf(deltaData.Slice(i * 2));
        var b = HalfTypeHelper.Convert(BitConverter.ToUInt16(deltaData.Slice(i * 2)));
        if (a != b)
            Debugger.Break();
        return a;
    })
    .ToArray();
```
and `Debugger.Break()` was indeed not called.
`HalfTypeHelper.Convert` implementation seems just about right for IEEE 754 half-precision float and C# `Half` is exactly that, so I guess HalfTypeHelper could be entirely removed, as VRF already targets .NET 6.

moofMonkey:
> Perhaps this PR could be simplified to decode all the frames like it was before?

Well, there are 4 options to be fully thread-safe
1) Remove frame cache, and always decode animation frames without caching - reduces framerate, but everything will be thread-safe.
2) `FrameCache` and `GetFrame` are only accessed from `GetTransformsAtTime`, so we could just add mutex for the `FrameCache` if multithreading is a concern
3) Precompute all frames, but that negates most of the benefits of the rewrite
4) Move all of `GetAnimationMatricesAsArray`/`GetAnimationMatrices`/`GetTransformsAtTime` to the GUI

moofMonkey:
I added `lock (FrameCache)`, I guess that's enough, but you can still choose other options.

xPaw:
3 is what I was thinking, and I don't think that negates any of the rewrite. You can combine that with option 1, removing the cache, and just adding the Frames array back which will decode all of them at once.

That should make it easier to reason about, then the optimizations can be considered in a separate smaller PR.

moofMonkey:
That's ok, I guess.
Peak RAM is still around the same value, however overall RAM usage is higher as it was with the old code.
I guess improved segment decoding that I didn't revert is what gives faster loading and less peak RAM usage, and deferring frame decoding is what gives lower overall RAM usage.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/SegmentDecoders/CCompressedStaticVector3.cs
xPaw: Yeah, see #287.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
```diff
@@ -52,7 +163,7 @@ public static IEnumerable<Animation> FromData(IKeyValueCollection animationData,
                 // Obviously we want to properly support animation decoding, but we are not there yet.
                 try
```
xPaw, line 164: This try/catch and the comment can be removed too now.

moofMonkey: Done, replaced with LINQ `.Select().ToArray()`

moofMonkey:
I cleaned up unneded `ref`s, there are a few more in [`NTRO.cs`](https://github.com/SteamDatabase/ValveResourceFormat/blob/master/ValveResourceFormat/Resource/ResourceTypes/NTRO.cs) but I guess it's out of scope for this PR, as it's only targeted at animations.
I initially used `ref` because `ReadSegment` accepted `ref`, even though it didn't need to.
Changed code works just as well without them.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/AnimationSegmentDecoder.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public AnimationSegmentDecoder(int[] elements, AnimationDataChannel localChannel)
+        {
+            Elements = elements;
+            LocalChannel = localChannel;
+        }
+
+        public virtual void Read(int frameIndex, Frame outFrame)
+        {
+            throw new NotImplementedException();
```
Perryvw, line 18: Why throw at runtime when you can just make this abstract and error at compile time

Perryvw (approved):

xPaw:
Thank you!

PR #465, opened 2022-12-08
glTF materials: metal texture & merge some logic with MaterialExtract

kristiker:
* Switched some logic in glTF material exporter to use sequences of `Texture*` parameters provided by `MaterialExtract` instead of raw `g_t` textures. This way any channel data that is not expected by the [glTF spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_material_pbrmetallicroughness_basecolortexture) can be stripped away or extracted separately (example: g_tColor alpha).

* Added CLI option `--gltf_textures_adapt` enabled by default in GUI. This option lets VRF perform operations on textures to make them match the glTF PBR spec (e.g. generate a metallicRoughness PBR texture & erase any useless channels).
  * Disabled 
image
  * Enabled
image

    * Disabled
    

    * Enabled
    

    * ^ Dota 2 assets export the inverted specular as roughness. A better way would be to use the specular workflow shader instead of metallic but seems like it's deprecated or something. I can't set it up.
        * By using the `MaterialExtract` shader table, this data was able to be extracted from `g_tMasks1` and `g_tMasks2` without additional hard-code.
    * Assets using the simple/complex shader
    
    

* Textures useless to glTF renderer are ignored for export (e.g. detail, masks1). However channels in these textures may still be collected if the `--gltf_textures_adapt` option is enabled.

### DIFF
--- a/ValveResourceFormat/IO/GltfModelExporter.cs
+++ b/ValveResourceFormat/IO/GltfModelExporter.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Numerics;
 using SharpGLTF.IO;
 using SharpGLTF.Schema2;
-using SkiaSharp;
 using ValveResourceFormat.Blocks;
 using ValveResourceFormat.ResourceTypes.ModelAnimation;
 using ValveResourceFormat.Serialization;
@@ -45,7 +44,6 @@ namespace ValveResourceFormat.IO
 
         private string DstDir;
         private readonly IDictionary<string, Node> LoadedUnskinnedMeshDictionary = new Dictionary<string, Node>();
-
 
         public static bool CanExport(Resource resource)
             => ResourceTypesThatAreGltfExportable.Contains(resource.ResourceType);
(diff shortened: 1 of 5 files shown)


ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -795,114 +866,139 @@ private Material GenerateGLTFMaterialFromRenderMaterial(VMaterial renderMaterial
...
+                ProgressReporter?.Report($"Exporting texture: {textureResource.FileName}");
+                var fileName = Path.GetFileName(textureResource.FileName);
+                var exportedTexturePath = Path.ChangeExtension(Path.Join(DstDir, fileName), "png");
 
-                var fileName = Path.GetFileNameWithoutExtension(texturePath);
+                using var bitmap = ((ResourceTypes.Texture)textureResource.DataBlock).GenerateBitmap();
+                using var pixels = bitmap.PeekPixels();
 
-                ProgressReporter?.Report($"Exporting texture: {texturePath}");
+                var gltfSlot = null as string;
```
xPaw, line 883: `string gltfSlot = null;`

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -752,6 +751,73 @@ private IEnumerable<(Node Node, List<int> Indices)> CreateBonesRecursive(Bone bo
                 .Append((node, bone.SkinIndices));
         }
 
+        /// <summary>
+        /// Packs metal and rough maps to metallicRoughnessTexture
+        /// </summary>
+        public class MRTextureGenerator : IDisposable
```
xPaw, line 757:
```suggestion
        public class MetallicRoughnesTextureGenerator : IDisposable
```

Is this class not useful when extracting materials outside of gltf?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -795,114 +866,139 @@ private Material GenerateGLTFMaterialFromRenderMaterial(VMaterial renderMaterial
...
+            {
+                var texturePath = renderTexture.Value;
+                var textureResource = FileLoader.LoadFile(texturePath + "_c");
+
+                if (textureResource == null)
+                {
+                    continue;
+                }
+
+                // TODO: get the signature directly instead of forming UnpackInfos
```
xPaw, line 972: todo.

The ToValueTuple seems rather pointless here.

ValveResourceFormat/IO/TextureExtract.cs
```diff
@@ -190,6 +191,56 @@ public static byte[] ToPngImageChannels(SKBitmap bitmap, Channel channel)
...
+
+        var srcPixelSpan = srcPixels.GetPixelSpan<SKColor>();
+        var pixelSpan = dstPixels.GetPixelSpan<SKColor>();
+
+        var inv = (byte)(invert ? 0xFF : 0x00);
+
+#pragma warning disable CS8509 // non exhaustive switch
+        for (var i = 0; i < srcPixelSpan.Length; i++)
+        {
+            pixelSpan[i] = dstChannel switch
```
xPaw, line 209: There's gotta be a simpler way of doing this, but I'm not sure.

ValveResourceFormat/IO/MaterialExtract.cs
```diff
@@ -147,6 +149,15 @@ public enum Channel
         }
     };
 
+    public static readonly Dictionary<string, ChannelMapping[]> GltfTextureMappings = new()
+    {
+        ["BaseColor"] = new[] { (Channel.RGB, "TextureColor"), (Channel.A, "TextureTranslucency") },
+        ["MetallicRoughness"] = new[] { (Channel.G, "TextureRoughness"), (Channel.B, "TextureMetalness") },
```
xPaw, line 155: How does this play along with MRTextureGenerator?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -795,114 +866,139 @@ private Material GenerateGLTFMaterialFromRenderMaterial(VMaterial renderMaterial
...
-                    case "g_tColorA":
-                    case "g_tColorB":
-                    case "g_tColorC":
-                        var channel = material.FindChannel("BaseColor");
-                        if (channel?.Texture != null && renderTexture.Key != "g_tColor")
-                        {
-                            break;
-                        }
+                    var image = model.UseImage(exportedTexturePath);
+                    image.Name = fileName + $"_{model.LogicalImages.Count - 1}";
```
xPaw, line 942: Put fileName in the interpolation while you're here

ValveResourceFormat/IO/GltfModelExporter.cs
kristiker: Not right now.

ValveResourceFormat/IO/MaterialExtract.cs
kristiker: For the `MetallicRoughness` entry, if there happens to be a source2 shader that defines a texture with same exact inputs as above then the generator will be skipped (desired).

However there isn't such shader in source2 at this time, and a more realistic set of inputs would occupy the red channel too, as even the glTF spec recommends putting Occlusion/Emmisive in there (or in alpha). I might need to make it "pattern match" for such scenarios (source2 {R, G, B} -> gltf {G, B}). Otherwise it's just there for consistency.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -839,113 +843,168 @@ private Material GenerateGLTFMaterialFromRenderMaterial(VMaterial renderMaterial
             }
 
             material.WithPBRMetallicRoughness(baseColor, null, metallicFactor: metalValue);
+            using var ormTexture = new TextureExtract.TexturePacker { DefaultColor = new SkiaSharp.SKColor(255, 255, 0, 255) };
+            var ormHasAO = false;
```
xPaw, line 847: ormHasAmbientOcclusion

And please add a comment above saying orm is Occlusion, Roughness, Metallic.

kristiker: Should be more clearer now.

Issue #379, opened 2021-09-22
Huge model files export into too many glTF bin files


hrimthurs:
### Describe the bug

When trying to decompile the model to gltf hangs at the stage "Found existing material" and generate many .bin files in out folder. Same result in gui and cli. On release version and unstable
![scr1](https://user-images.githubusercontent.com/65903670/134350295-c1c7eecf-04fd-45ed-a58d-27987aceaa2b.JPG)
![scr2](https://user-images.githubusercontent.com/65903670/134350330-890155ed-c351-4919-ab73-78dc4d9fa3d5.JPG)

.

### What Valve game does this happen in

Dota 2

### Reference file in a Source 2 game

models/items/spectre/spectre_arcana/spectre_arcana_base

### Expected behavior

Decompilated .gltf in out folder. As with other models.

### What version of VRF are you using? On what platform?

VRF 0.1.10.828, Win10 x64

xPaw:
It's slow, have you waited long enough?

Can you press cancel to see anything in console?

faust-hub:
I waited until there were too many files (120000 files summary 180Mb). Progress bar is standing
This only happened on this model.
After canceling in the console:
![scr3](https://user-images.githubusercontent.com/65903670/134354845-17ca0b3f-107e-40d8-9658-f502a8b97266.JPG)

xPaw:
There is a limitation with gltf where too big models won't fit into one file, so it's being split into many small ones. It's not ideal, but from what I'm seeing it's not stuck, it's just taking a long time to export.

faust-hub:
You're right. This is not a bug, but a very bad model. I waited a very long time to unpack it and got the result of an incredibly large size. The VRF works great. Valve sucks

Teajai2:
I attempted to decompile the Half-Life Alyx zombie and leave the process running overnight. When I woke up, my computer was incredibly slow, so I restarted my PC and I could not get it to boot it up again.

As it would turn out, my hard drive broke, and I had to literally get a new one in order to continue using my computer.

PR #467, opened 2022-12-10
glTF improvements

moofMonkey:
This PR fixes
- Animations being duplicated in resulting glTF which resulted in huge glTF size - by reusing the same skeleton
- Animations related errors in resulting glTF caused by mixing `WithLocalTransform` and `WithLocalTranslation`/`WithLocalRotation`
- Int32 overflow check in WriteModelFile

and adds animation channel data deduplication.

Fixes most, if not all cases of #379.
Dota 2 `models/items/spectre/spectre_arcana/spectre_arcana_base.vmdl` mentioned in the issue gets perfectly exported with a more or less reasonable size of 75.5MB (instead of 1GB+).
sandbox babylonjs com_

### DIFF
--- a/ValveResourceFormat/IO/GltfModelExporter.cs
+++ b/ValveResourceFormat/IO/GltfModelExporter.cs
@@ -383,36 +383,28 @@ namespace ValveResourceFormat.IO
         /// <returns>A tuple of meshes and their names.</returns>
         private (VMesh Mesh, string Name)[] LoadModelMeshes(VModel model)
         {
-            var refMeshes = model.GetRefMeshes().ToArray();
-            var meshes = new (VMesh, string)[refMeshes.Length];
+            var embeddedMeshes = model.GetEmbeddedMeshesAndLoD()
+                .Where(m => (m.LoDMask & 1) != 0)
+                .Select((m, i) => (m.Mesh, $"Embedded.{i}"));
 
-            var embeddedMeshIndex = 0;
-            var embeddedMeshes = model.GetEmbeddedMeshes().ToArray();
-
-            for (var i = 0; i < meshes.Length; i++)
-            {
-                var meshReference = refMeshes[i];
-                if (string.IsNullOrEmpty(meshReference))
-                {
-                    // If refmesh is null, take an embedded mesh
-                    meshes[i] = (embeddedMeshes[embeddedMeshIndex++], $"Embedded.{embeddedMeshIndex}");
-                }
-                else
+            var refMeshes = model.GetReferenceMeshNamesAndLoD()
+                .Where(m => (m.LoDMask & 1) != 0)
+                .Select(m =>
                 {
                     // Load mesh from file
-                    var meshResource = FileLoader.LoadFile(meshReference + "_c");
+                    var meshResource = FileLoader.LoadFile(m.MeshName + "_c");
+                    var nodeName = Path.GetFileNameWithoutExtension(m.MeshName);
                     if (meshResource == null)
                     {
-                        continue;
+                        return (null, nodeName);
                     }
 
-                    var nodeName = Path.GetFileNameWithoutExtension(meshReference);
                     var mesh = new VMesh(meshResource);
-                    meshes[i] = (mesh, nodeName);
-                }
-            }
+                    return (mesh, nodeName);
+                })
+                .Where(m => m.mesh != null);
 
-            return meshes;
+            return embeddedMeshes.Concat(refMeshes).ToArray();
         }
 
         /// <summary>
--- a/ValveResourceFormat/IO/GltfModelExporter.cs
+++ b/ValveResourceFormat/IO/GltfModelExporter.cs
@@ -7,7 +7,6 @@ using SharpGLTF.IO;
 using SharpGLTF.Schema2;
 using SkiaSharp;
 using ValveResourceFormat.Blocks;
-using ValveResourceFormat.ResourceTypes;
 using ValveResourceFormat.ResourceTypes.ModelAnimation;
 using ValveResourceFormat.Serialization;
 using ValveResourceFormat.Utils;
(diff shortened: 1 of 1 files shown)


moofMonkey:
I'll add logic to omit static data, as in my test it omits ~30% of resulting model size (126MB result)

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -346,7 +359,7 @@ public void ExportToFile(string resourceName, string fileName, VModel model)
                 if (string.IsNullOrEmpty(meshReference))
                 {
                     // If refmesh is null, take an embedded mesh
-                    meshes[i] = (embeddedMeshes[embeddedMeshIndex++], $"{modelName}.Embedded.{embeddedMeshIndex}");
```
Perryvw, line 349: Just checking: Does this not cause name collisions in files with multiple models with embedded meshes?

moofMonkey: As far as I can tell, glTF doesn't really care about name collisions.

Either way, I removed `modelName` from embedded meshes names due to models now having their own nodes - this way there are less `WorldMatrix`es in resulting glTF, and it's less bloated when there are multiple models present.
https://github.com/SteamDatabase/ValveResourceFormat/pull/467/files#diff-a740f76f43bd153e9126ae3abe86209608d28825f381ed37b2d74aa3bcbc9c93R278
image

moofMonkey:
181MB => 75.5MB. That's pretty nice, and keyframes should be identical.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -309,21 +266,110 @@ public void ExportToFile(string resourceName, string fileName, VModel model)
...
+                            var bone = boneFrame.Key;
+                            if (!rotationDict.ContainsKey(bone))
+                            {
+                                rotationDict[bone] = new Dictionary<float, Quaternion>();
+                                translationDict[bone] = new Dictionary<float, Vector3>();
+                            }
+
+                            if (!lastRotationDict.TryGetValue(bone, out var lastRotation) || lastRotation != boneFrame.Value.Angle)
+                            {
+                                if (rotationOmittedSet.Contains(bone))
```
xPaw, line 310: Use Remove directly, it returns a bool.

moofMonkey:
Added comments for `rotationOmittedSet`/`translationOmittedSet` and fixed `GetGltfSkeletonJoints` logic for `numJoints`.

moofMonkey:
Fixed exception in case where Skeleton#Bones contained `null` for some reason - only applicable to this PR, as I removed recursion when retrieving joints.

Whole Dota 2 `models/creeps` folder gets exported without exceptions now.

Perryvw (approved):

moofMonkey:
Probably fixed exception when model didn't have meshes and skeletons - happened when I tried to export Dota 2 `maps/dota.vpk` `maps/dota/world.vwrld` as glTF.

Fixed standalone meshes not being transformed because WorldMatrix was set for `skeletonNode`, and not `modelNode`, and only models have `skeletonNode`.

Fixed LoD != 0 being exported to glTF.

xPaw:
> I tried to remove exporting of LoD != 0, but apparently it breaks export of some models like Dota 2 models/items/spectre/spectre_arcana/spectre_arcana_base.vmdl.

Does it break in the gui viewer too?

kristiker:
>I tried to remove exporting of LoD != 0

I was thinking of adding this extension: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/MSFT_lod. Looks like the engine that you're using supports it. Although an option to ignore them completely would be useful too.

moofMonkey:
> Does it break in the gui viewer too?

No, the GUI viewer loads meshes differently and I copied code from there, which is why it didn't work.
It was breaking because of `LoadModelMeshes` `string.IsNullOrEmpty(meshReference)` being incompatible with my first try. Basically embedded meshes didn't load at all.
Reimplemented `LoadModelMeshes` and now it doesn't export LoD != 0 and works properly.

> I was thinking of adding this extension: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/MSFT_lod. > Looks like the engine that you're using supports it.

At this time SharpGLTF [lacks support](https://github.com/vpenades/SharpGLTF/blob/master/src/SharpGLTF.Core/README.md#:~:text=MSFT_texture_dds-,MSFT_lod,-When%20this%20extension) for MSFT_lod.

xPaw (approved):

kristiker:
It's not working right when performing a folder export like `dota/models/architecture`.

moofMonkey:
> It's not working right when performing a folder export like dota/models/architecture. Error:
> Failed to extract 'models/architecture/continuum/continuum_ground_01_crn_02.vmdl_c' - LogicalParent mismatch (Parameter 'value')

Sorry, can't reproduce errors on this specific file nor folder at 08c567ba69d0bd0365d665ab858be39276ef1171. What commit are you at?
Model is being exported without any errors, and properly opens at sandbox.babylonjs.com.
sandbox babylonjs com_ (1)

> Single models though export so much faster now 💯. Yesterday this model pulled 5GB of ram to export, now it goes super quick!

Glad to hear that, hope it's easier to make your PR now.

xPaw:
I get these errors too when exporting a *folder*.

moofMonkey:
Can confirm, sorry, didn't check the console tab while exporting folders.

Will investigate and submit yet another PR if it doesn't get fixed before that, I guess.

moofMonkey:
The problem is in `LoadedUnskinnedMeshDictionary` being keyed by the mesh name, which was changed in this PR to not include model name for embedded meshes.

There won't be any problems with the Decompiler, but GUI for some reason reuses single instance of `GltfModelExporter`, and that causes meshes from different files to mix up. Guess this PR just revealed an existing problem.

kristiker:
Should clear the dictionary after every LoadModel then I think (and also on the Vmesh case that doesn't call LoadModel).

Issue #470, opened 2022-12-10
Canceling export in the export dialog does not actually cancel the exporting


xPaw:
It only hides the dialog.

PR #471, opened 2022-12-10
Fix glTF folder export

moofMonkey:
This PR fixes
- Mesh cache not being cleared after glTF export
- GetSkeleton throwing on invalid meshIndex (to avoid workarounds in other code)
- glTF `NODE_SKINNED_MESH_NON_ROOT` warning by removing behavior introduced in #467 - otherwise mesh reusing is not possible due to LogicalParent check

and allows much better mesh reusing via `LoadedMeshDictionary`, this way export of Dota 2 `maps/dota_728.vpk` `maps/dota/world.vwrld` is now possible within a reasonable time (2m22s300ms on my PC from GUI).
Can't check Dota 2 `maps/dota.vpk` export because it throws because of vphys parsing right now.

Also I don't know whether that's old or new behavior, but exported dota map materials aren't working properly.
sandbox babylonjs com_

### DIFF
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -844,7 +844,7 @@ namespace Decompiler
 
                             Directory.CreateDirectory(Path.GetDirectoryName(outputFile));
 
-                            gltfModelExporter.Export(resource, outputFile);
+                            gltfModelExporter.Export(resource, outputFile, null);
 
                             continue;
                         }
--- a/Decompiler/Decompiler.cs
+++ b/Decompiler/Decompiler.cs
@@ -788,6 +788,14 @@ namespace Decompiler
             }
 
             var fileLoader = new BasicVpkFileLoader(package);
+            var gltfModelExporter = new GltfModelExporter
+            {
+                ExportMaterials = GltfExportMaterials,
+                AdaptTextures = GltfExportAdaptTextures,
+                ProgressReporter = new Progress<string>(progress => Console.WriteLine($"--- {progress}")),
+                FileLoader = fileLoader
+            };
+
             var entries = package.Entries[type];
 
             foreach (var file in entries)
@@ -835,14 +843,6 @@ namespace Decompiler
                             var outputFile = Path.Combine(OutputFile, Path.ChangeExtension(filePath, outputExtension));
 
                             Directory.CreateDirectory(Path.GetDirectoryName(outputFile));
-
-                            var gltfModelExporter = new GltfModelExporter
-                            {
-                                ExportMaterials = GltfExportMaterials,
-                                AdaptTextures = GltfExportAdaptTextures,
-                                ProgressReporter = new Progress<string>(progress => Console.WriteLine($"--- {progress}")),
-                                FileLoader = fileLoader
-                            };
 
                             gltfModelExporter.Export(resource, outputFile);
 
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -180,7 +180,8 @@ namespace GUI.Forms
 
                             if (GltfModelExporter.CanExport(resource))
                             {
-                                gltfExporter.Export(resource, Path.ChangeExtension(outFilePath, "glb"));
+                                gltfExporter.Export(resource, Path.ChangeExtension(outFilePath, "glb"),
+                                    cancellationTokenSource.Token);
                                 continue;
                             }
 
@@ -246,6 +247,7 @@ namespace GUI.Forms
         {
             foreach (var contentSubFile in contentFile.SubFiles)
             {
+                cancellationTokenSource.Token.ThrowIfCancellationRequested();
                 contentSubFile.FileName = Path.Combine(contentRelativeFolder, contentSubFile.FileName).Replace(Path.DirectorySeparatorChar, '/');
                 var fullPath = Path.Combine(path, contentSubFile.FileName);
 
--- a/ValveResourceFormat/IO/GltfModelExporter.cs
+++ b/ValveResourceFormat/IO/GltfModelExporter.cs
@@ -45,7 +45,6 @@ namespace ValveResourceFormat.IO
         public bool AdaptTextures { get; set; } = true;
 
         private string DstDir;
-        private readonly IDictionary<string, Mesh> LoadedMeshDictionary = new Dictionary<string, Mesh>();
         private CancellationToken? CancellationToken;
 
         public static bool CanExport(Resource resource)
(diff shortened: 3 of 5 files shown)


kristiker (commented):
> GUI reusing GltfModelExporter, which is only intended to be used for one Scene because of mesh caching

Aren't meshes a per-model thing? Shouldn't these meshes be cleared after each specific model in a worldnode scene?

moofMonkey:
> Aren't meshes a per-model thing? Shouldn't these meshes be cleared after each specific model in a worldnode scene?

Usually meshes get loaded from models, yes, but they can be reused in glTF as long as they're in the same Scene, aka export file.

Mostly this impacts exported worlds, but there might be some models that use the same mesh multiple times.

Decompiler/Decompiler.cs
```diff
@@ -788,6 +788,14 @@ private void DumpVPK(string parentPath, Package package, string type, Dictionary
             }
 
             var fileLoader = new BasicVpkFileLoader(package);
+            var gltfModelExporter = new GltfModelExporter
```
xPaw, line 791: You can specify multiple threads in decompiler, and gltf exporter isnt thread safe, so it's very likely that this would cause issues. Keep it as it currently is.

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -246,6 +246,7 @@ private async Task ExtractSubfiles(string contentRelativeFolder, ContentFile con
         {
             foreach (var contentSubFile in contentFile.SubFiles)
             {
+                cancellationTokenSource.Token.ThrowIfCancellationRequested();
```
xPaw, line 249: FWIW a good place to add one of these, but doesn't fully fix the issue. Need to pass in the cancellation token into gltf exporter itself  and place a few throws in there too. Because exporting something huge will still take a very long time (e.g. map adding many models, or writing textures to disk).

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -44,29 +44,36 @@ public class GltfModelExporter
...
+                    case ResourceType.World:
+                        ExportToFile(resource.FileName, targetPath, (VWorld)resource.DataBlock);
+                        break;
+                    default:
+                        throw new ArgumentException($"{resource.ResourceType} not supported for gltf export");
+                }
+            }
+            finally
+            {
+                LoadedMeshDictionary.Clear();
```
xPaw, line 76: Other ExportToFile overloads are public too, and it won't be cleared if consumer calls them directly instead of this method.

Decompiler/Decompiler.cs
moofMonkey: This specific function - DumpVPK - is not affected by threads setting, and furthermore it shouldn't be, as it'd race condition against itself on exporting same files.

In my tests of making `foreach (var file in entries)` multithreaded via queue (not committed anywhere) speeds it up from ~10m to ~6m for Dota 2 `-d -f models/ -e vmdl_c`, but the question of how to make it not race condition would still be open.

xPaw: > is not affected by threads setting
it is if you follow the call graph up to ParseVPK and then ProcessFile.

Exporter has DstDir which makes it not thread safe.

GUI/Forms/ExtractProgressForm.cs
moofMonkey: > Need to pass in the cancellation token into gltf exporter itself and place a few throws in there too.

Done, but ExportFile#Export has no cancellation token to pass to glTF exporter, so exporting single file (i.e. world) will still not cancel properly.

ValveResourceFormat/IO/GltfModelExporter.cs
moofMonkey: Well then the only adequate solution left is passing cache as argument. Done.

Decompiler/Decompiler.cs
moofMonkey: > it is if you follow the call graph up to ParseVPK and then ProcessFile.

That path doesn't touch DumpVPK as far as I can tell.

moofMonkey: Anyway this can be resolved as `Don't reuse cache between glTF exports` makes glTF exporter thread-safe.

xPaw: Since you made the cache local, probably better to pass it as an argument, so at least that solves this particular thread safety.

> That path doesn't call DumpVPK as far as I can tell
I think it is.

> glTF exporter inside DumpVPK is not passed to any other threads whatsoever.
This you're right about, as it's local.

GUI/Forms/ExtractProgressForm.cs
xPaw: That method should probably be updated to use ExtractProgressForm for consistency, but that can be done in a future PR.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -76,7 +76,7 @@ public void Export(Resource resource, string targetPath)
         /// <param name="resourceName">The name of the resource being exported.</param>
         /// <param name="fileName">Target file name.</param>
         /// <param name="world">The world resource to export.</param>
-        public void ExportToFile(string resourceName, string fileName, VWorld world)
+        public void ExportToFile(string resourceName, string fileName, VWorld world, CancellationToken? cancellationToken)
```
xPaw, line 79: Need to add an xmldoc.

Seems like `public void Export` is missing an xmldoc entirely too.

moofMonkey: Done.

Decompiler/Decompiler.cs
moofMonkey: > Since you made the cache local, probably better to pass it as an argument

It was already only applicable within a single exported file, and would break if, for example, one mesh was referenced in different files in a folder.
It only applied to meshes without skeleton before those changes though, which probably just hid the problem for the most part.

Anyway yes, now it is passed as an argument now to avoid try/catches.

GUI/Forms/ExtractProgressForm.cs
moofMonkey: Yeah I don't really want to dig into GUI code, so I'd leave it for someone else. 

Removed `Fixes #470` from commit description, also sorry for the reference spam caused by force-pushes.

moofMonkey: Easiest path to have cancellation token there would be making `GenericProgressForm` `cancellationTokenSource` public.

PR #474, opened 2022-12-11
Update FPS counter once a second

xPaw:
- Slightly optimize fps display
- Update FPS counter once a second, count frames manually

### DIFF
--- a/GUI/Controls/GLViewerControl.cs
+++ b/GUI/Controls/GLViewerControl.cs
@@ -65,9 +65,9 @@ namespace GUI.Controls
             glControlContainer.Controls.Add(GLControl);
         }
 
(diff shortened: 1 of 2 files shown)


GUI/Controls/GLViewerControl.cs
```diff
@@ -16,6 +16,9 @@ namespace GUI.Controls
 {
     internal partial class GLViewerControl : UserControl
     {
+        private const long TicksPerSecond = 10_000_000;
```
Perryvw, line 19: Would probably be nicer to define TicksPerSecond as a function of the frequency and the stopwatch frequency, instead of the other way around

Perryvw (approved):

PR #478, opened 2022-12-12
Use ExtractProgressForm for single file exports

xPaw:
- Move extract methods to ExportFile
- Make single file export use ExtractProgressForm
- Fixed folder export not using gltf progress reporter


GUI/Forms/ExtractProgressForm.cs
```diff
@@ -49,60 +47,65 @@ public ExtractProgressForm(Package package, TreeNode root, string path, ExportDa
...
                         foreach (var resourceType in ExtractOrder)
                         {
-                            Invoke(() => Text = $"Extracting {resourceType}s...");
+                            Invoke(() => extractStatusLabel.Text = $"Extracting {resourceType}s...");
                             var extension = FileExtract.GetExtension(resourceType);
                             await ExtractFilesAsync(filesToExtractSorted[extension + "_c"]).ConfigureAwait(false);
                         }
 
-                        Invoke(() => Text = $"Extracting files...");
+                        Invoke(() => extractStatusLabel.Text = $"Extracting files...");
```
kristiker, line 93: The subsequent `ExtractFilesAsync` immediately replaces these messages. Should keep them on title. When running a materials folder the title should say "Extracting Materials...", then "Extracting files..." when it moves to anything non-vmat. Also the "Calculating..." above doesn't seem to calculate anymore.

xPaw: I changed it to a log window.

PR #479, opened 2022-12-12
Implement deferred animations decoding

moofMonkey:
Should greatly reduce memory usage by both GUI and decompiler

`GetFrame` (previously `GetTransformsAtTime`) inclusive CPU time spent on `models/items/spectre/spectre_arcana/spectre_arcana_base.vmdl_c` on `spct_arc_flail` animation in GUI in Release mode at ~120FPS:
without previous/next frame cache: ~29.43%
with previous/next frame cache: ~5.63%
with frame cache, but without reusing `FrameBone`s: ~6.17%

For comparison on master (48a28406a84c137b5bb914e3cefff856ddb83be9) - frames are always cached, but interpolated frame/bones aren't reused: ~5.58%

It also uses ~350-400MB less RAM than on master in GUI in Release mode on given model and animation.

### DIFF
--- a/GUI/Types/Renderer/WorldLoader.cs
+++ b/GUI/Types/Renderer/WorldLoader.cs
@@ -273,7 +273,7 @@ namespace GUI.Types.Renderer
 
                 if (animation != default)
                 {
-                    modelNode.LoadAnimation(animation); // Load only this animation
+                    modelNode.LoadAnimations();
                     modelNode.SetAnimation(animation);
                 }
 
--- a/GUI/Types/Renderer/WorldLoader.cs
+++ b/GUI/Types/Renderer/WorldLoader.cs
@@ -152,8 +152,7 @@ namespace GUI.Types.Renderer
                 var model = entity.GetProperty<string>("model");
                 var skin = entity.GetProperty<string>("skin");
                 var particle = entity.GetProperty<string>("effect_name");
-                //var animation = entity.GetProperty<string>("defaultanim");
-                string animation = null;
+                var animation = entity.GetProperty<string>("defaultanim");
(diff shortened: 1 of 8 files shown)


GUI/Types/Renderer/WorldLoader.cs
```diff
@@ -274,7 +274,7 @@ private void LoadEntitiesFromLump(Scene scene, LoadResult result, EntityLump ent
 
                 if (animation != default)
                 {
-                    modelNode.LoadAnimation(animation); // Load only this animation
+                    modelNode.LoadAnimations(guiContext.FileLoader);
```
xPaw, line 277: Won't this load every single animation file unnecessarily, which this was designed to avoid. Have you compared this to master? Dota map might not be a good enough test, would need a map that has models with many animations.

moofMonkey: Is it actually worth it, especially given that code path is unused right now?

It loaded all embedded animations either way, and if animation wasn't present there - it loaded all referenced animation groups until needed animation was found, so in the worst case (missing animation or it's present in the last file) it'd load all referenced animations as well.

The worst is that it only added reference animations from file that contained needed animation, so if that path triggered multiple times for ref animations - it loaded them over and over.
https://github.com/SteamDatabase/ValveResourceFormat/blob/f06c48d3f48d959b3a79ae43e7226a39e27d4cb1/GUI/Types/Renderer/ModelSceneNode.cs#L246-L250

xPaw: I tested on a2_drainage, looks like it's not a problem.

PR #484, opened 2022-12-14
Gltf memory and image overwriting fixes

kristiker:
In satellite images mode, [SharpGLTF](https://github.com/vpenades/SharpGLTF)'s `model.UseImage` will read the file from disk and hold data in memory so it can do the writing itself. This operation is totally unnecessary when VRF has just written the image to disk.

Due to API limitations I had to write a 44 byte png, send that image to SharpGLTF for linking like before, and then replace the 44 byte png on disk with the full image. Then through the provided callback, SharpGLTF's image writing is canceled the URI is fixed up. This should reduce memory consumption.
**Edit:** After an API update on SharpGLTF' side, writing the empty image on disk is not needed.

The sum of images will be kept in memory only when using embedded image mode.

### DIFF
--- a/ValveResourceFormat/IO/GltfModelExporter.cs
+++ b/ValveResourceFormat/IO/GltfModelExporter.cs
@@ -53,7 +53,9 @@ namespace ValveResourceFormat.IO
         private string DstDir;
         private CancellationToken? CancellationToken;
 
-        private readonly byte[] tempPng = new byte[] { 137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82, 0, 0, 0, 1, 0, 0, 0, 1, 8, 6, 0, 0, 0, 255, 255, 255, 255, 255, 255, 255, 73, 69, 78, 68, 174, 66, 96, 130 };
+        // In SatelliteImages mode, SharpGLTF will still load and validate images.
+        // To save memory, we initiate MemoryImage with a a dummy image instead.
+        private readonly byte[] dummyPng = new byte[] { 137, 80, 78, 71, 0, 0, 0, 0, 0, 0, 0, 0 };
 
         public static bool CanExport(Resource resource)
             => ResourceTypesThatAreGltfExportable.Contains(resource.ResourceType);
@@ -1107,6 +1109,7 @@ namespace ValveResourceFormat.IO
 
             using var occlusionRoughnessMetal = new TextureExtract.TexturePacker { DefaultColor = new SkiaSharp.SKColor(255, 255, 0, 255) };
             var ormHasOcclusion = false;
+            Image ormImage = null;
 
             var allGltfInputs = MaterialExtract.GltfTextureMappings.Values.SelectMany(x => x);
             var blendNameComparer = new MaterialExtract.LayeredTextureNameComparer(new HashSet<string>(allGltfInputs.Select(x => x.Item2)));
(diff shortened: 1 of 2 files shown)


xPaw:
Does the png need to be valid? Is it not possible to avoid making sharpgltf read it?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -1055,6 +1023,50 @@ void WriteTexture(MaterialExtract.Channel channel, string gltfName, int index, i
...
+                image = model.CreateImage(fileName);
+                image.Content = TextureExtract.ToPngImageChannels(bitmap, channel);
+                return image;
+            }
+
+            var exportedTexturePath = Path.ChangeExtension(Path.Join(DstDir, fileName), "png");
+
+            // Write a 44 byte PNG file that can be replaced with the full image after linking to modelRoot.
+            // SharpGLTF loads images into memory, which isn't necessary with satellite images.
+            using (var fs = File.Open(exportedTexturePath, FileMode.Create))
```
xPaw, line 1048: You can keep the fs open, rewind it and then write the full image.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -530,7 +533,7 @@ private void WriteModelFile(ModelRoot exportedModel, string filePath)
 
             var settings = new WriteSettings
             {
-                ImageWriting = ResourceWriteMode.SatelliteFile,
+                ImageWriting = SatelliteImages ? ResourceWriteMode.SatelliteFile : ResourceWriteMode.BufferView,
                 ImageWriteCallback = ImageWriteCallback,
```
xPaw, line 537: Just for correctness sake, is ImageWriteCallback called when using ResourceWriteMode.BufferView?

kristiker: It's called in `ResourceWriteMode.SatelliteFile` mode only.

ValveResourceFormat/IO/GltfModelExporter.cs
kristiker: The CreateImage call just below this needs the file to be free.

xPaw: What if you specify FileShare.Read?

kristiker: That's good idea but can't get it to work.. I tried
`File.Open(exportedTexturePath, FileMode.Create, FileAccess.Write, FileShare.Read)`
`File.Open(exportedTexturePath, FileMode.Create, FileAccess.Write, FileShare.Write)`

Still throwing IOException.

xPaw:
https://github.com/vpenades/SharpGLTF/commit/180c80308566d2bbe8144b32e0eddabaabed7a74

` Image.AlternateWriteFileName` was added which should fix the need to write an empty png to disk.

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -1055,6 +1023,44 @@ void WriteTexture(MaterialExtract.Channel channel, string gltfName, int index, i
             return material;
         }
 
+        private Image LinkAndStoreImage(MaterialExtract.Channel channel, SkiaSharp.SKBitmap bitmap, ModelRoot model, string fileName)
+        {
+            var existingImage = model.LogicalImages.SingleOrDefault(i => i.Name == fileName);
+            if (existingImage is not null)
+            {
+                return existingImage;
```
xPaw, line 1031: It seems like such a cache check should be done way earlier, before it even has a bitmap?

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -43,10 +43,13 @@ public class GltfModelExporter
         public IFileLoader FileLoader { get; set; }
         public bool ExportMaterials { get; set; } = true;
         public bool AdaptTextures { get; set; } = true;
+        public bool SatelliteImages { get; set; } = true;
 
         private string DstDir;
         private CancellationToken? CancellationToken;
 
+        private readonly byte[] tempPng = new byte[] { 137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82, 0, 0, 0, 1, 0, 0, 0, 1, 8, 6, 0, 0, 0, 255, 255, 255, 255, 255, 255, 255, 73, 69, 78, 68, 174, 66, 96, 130 };
```
xPaw, line 51: Add a comment explaining why this is required

kristiker: Done.

ValveResourceFormat/IO/GltfModelExporter.cs
kristiker: Fixed, bitmap is generated after this check. For both this and the orm texture.

xPaw:
There was sharpgltf update, so this can be updated and merged.

kristiker:
Seems to be working fine.

PR #486, opened 2022-12-15
Filter out children in glTF folder exports to avoid duplicates

### LINKED ISSUE #481: glTF folder export is not smart and overwrites files all the time

In Dota 2 folders may contain vmdl_c and vmesh_c that go by the same name. As far as I can tell vmdl_c is a superset of vmesh_c so the export should ideally prioritize vmdl_c.

Models (such as those contained by worldnodes ) often share texture images, these also get overwritten.

Skip exporting the same data twice, it will speed up exports. Iirc the vmat generator already handles this issue well.

_No response_

kristiker:
The order of exports for gltf-exportables is now
```py
ResourceType.World
ResourceType.WorldNode
ResourceType.Model
ResourceType.Mesh
ResourceType.AnimationGroup
ResourceType.Animation
ResourceType.Sequence
ResourceType.Morph
```

This means that when exporting folders in GUI, WorldNodes referenced by World, Meshes referenced by Models and so on will be skipped.

image
* This folder now exports 3 GLB files, previously it did 6.
* Some cases had the vmesh_c with same name as vmdl_c, which would result in an overwrite with potential loss of data since vmesh_c is a subset of vmdl_c.

Closes #481

### DIFF
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -26,7 +26,15 @@ namespace GUI.Forms
 
         private static readonly List<ResourceType> ExtractOrder = new()
         {
-            // Materials before textures
+            ResourceType.World,
+            ResourceType.WorldNode,
+            ResourceType.Model,
+            ResourceType.Mesh,
+            ResourceType.AnimationGroup,
+            ResourceType.Animation,
+            ResourceType.Sequence,
+            ResourceType.Morph,
+
             ResourceType.Material,
             ResourceType.Texture,
         };
@@ -56,7 +64,7 @@ namespace GUI.Forms
             {
                 gltfExporter = new GltfModelExporter()
                 {
-                    FileLoader = exportData.VrfGuiContext.FileLoader,
+                    FileLoader = new TrackingFileLoader(exportData.VrfGuiContext.FileLoader),
                     ProgressReporter = new Progress<string>(SetProgress),
                 };
             }
@@ -194,6 +202,11 @@ namespace GUI.Forms
             if (GltfModelExporter.CanExport(resource) && Path.GetExtension(outFilePath) is ".glb" or ".gltf")
             {
                 gltfExporter.Export(resource, outFilePath, cancellationTokenSource.Token);
+                if (gltfExporter.FileLoader is TrackingFileLoader trackingFileLoader)
+                {
+                    extractedFiles.UnionWith(trackingFileLoader.LoadedFilePaths);
+                }
+
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -278,18 +282,19 @@ public void ExportToFile(string resourceName, string fileName, VModel model, Can
...
+            LoadModel(exportedModel, scene, model, resourceName, Matrix4x4.Identity, loadedMeshDictionary, externalRefsHandled);
 
             WriteModelFile(exportedModel, fileName);
+            return externalRefsHandled;
         }
 
         private void LoadModel(ModelRoot exportedModel, Scene scene, VModel model, string name,
-            Matrix4x4 transform, IDictionary<string, Mesh> loadedMeshDictionary, string skinName = null)
+            Matrix4x4 transform, IDictionary<string, Mesh> loadedMeshDictionary, HashSet<string> externalRefsHandled, string skinName = null)
         {
```
kristiker, line 298: The method handles vanim_c and vagrp_c too but `model.GetAllAnimations(FileLoader)` hides their filenames out.

GUI/Forms/ExtractProgressForm.cs
```diff
@@ -26,7 +26,11 @@ public partial class ExtractProgressForm : Form
 
         private static readonly List<ResourceType> ExtractOrder = new()
         {
-            // Materials before textures
+            ResourceType.World,
+            ResourceType.WorldNode,
+            ResourceType.Model,
+            ResourceType.Mesh,
```
xPaw, line 32: I can Morph can be added to the list now, as master now parses it (and thus loads atlas texture).

ValveResourceFormat/IO/GltfModelExporter.cs
```diff
@@ -103,7 +103,8 @@ public void ExportToFile(string resourceName, string fileName, VWorld world, Can
                 {
                     continue;
                 }
-                var worldResource = FileLoader.LoadFile(worldNodeName + ".vwnod_c");
+
+                var worldResource = FileLoader.LoadFile(worldNodeName.Replace('\\', '/') + ".vwnod_c");
```
xPaw, line 107: This is a bit silly, file loader should probably handle this instead.

kristiker: I'm guessing the tracking one. Sure but it's only this case where the path is malformed.

xPaw: That we know of.

PR #496, opened 2022-12-25
Improve bones remapping

moofMonkey:
Improves animations performance of models with multiple meshes by O(MeshCount) by both calculating and uploading animation matrices only once per model, not per mesh.

Also fixes:
- Remap table being larger than model bone count;
- Most of glTF validation errors - i.e. `DUPLICATE_ELEMENTS` on `/joints/`.

Uses a bit more memory because of vertex buffer duplication, but that could be improved, I guess.
Difference for Aperture Desk Job map is just ~4.5% (4.5GB => 4.7GB).

### DIFF
--- a/ValveResourceFormat/Resource/Blocks/VBIB.cs
+++ b/ValveResourceFormat/Resource/Blocks/VBIB.cs
@@ -4,6 +4,8 @@ using System.IO;
 using System.Text;
 using ValveResourceFormat.Compression;
 using ValveResourceFormat.Serialization;
+using System.Linq;
+using System.Runtime.InteropServices;
 
 namespace ValveResourceFormat.Blocks
 {
@@ -24,7 +26,7 @@ namespace ValveResourceFormat.Blocks
             //stride for vertices. Type for indices
             public uint ElementSizeInBytes;
             //Vertex attribs. Empty for index buffers
-            public List<RenderInputLayoutField> InputLayoutFields;
+            public RenderInputLayoutField[] InputLayoutFields;
             public byte[] Data;
         }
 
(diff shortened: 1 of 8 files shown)


ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
```diff
@@ -152,19 +152,15 @@ private static IKeyValueCollection GetAnimationData(Resource resource)
         public Matrix4x4[] GetAnimationMatrices(AnimationFrameCache frameCache, float time, Skeleton skeleton)
         {
             // Create output array
-            var matrices = new Matrix4x4[skeleton.AnimationTextureSize];
-
-            for (var i = 0; i < skeleton.AnimationTextureSize; i++)
-            {
-                matrices[i] = Matrix4x4.Identity;
-            }
+            var matrices = Enumerable.Range(0, skeleton.Bones.Length)
```
xPaw, line 155: old loop was fine.

moofMonkey: I've been changing code a lot, right now it doesn't even require filling identity matrices.

Changed to plain `new Matrix4x4[skeleton.Bones.Length]`.

ValveResourceFormat/Resource/Blocks/VBIB.cs
```diff
@@ -353,5 +377,81 @@ public override void WriteText(IndentedTextWriter writer)
...
+            var res = new VBIB();
+            res.VertexBuffers.AddRange(VertexBuffers.Select(buf =>
+            {
+                var blendIndices = Array.FindIndex(buf.InputLayoutFields, field => field.SemanticName == "BLENDINDICES");
+                if (blendIndices != -1)
+                {
+                    var field = buf.InputLayoutFields[blendIndices];
+                    var (formatElementSize, formatElementCount) = GetFormatInfo(field);
+                    var formatSize = formatElementSize * formatElementCount;
+                    buf.Data = buf.Data.ToArray();
```
xPaw, line 426: This seems suspicious, aren't you modifying the original VBIB's data?

moofMonkey: Nope, OnDiskBufferData is a struct, so is passed by copy

PR #497, opened 2022-12-27
Optimize animations

moofMonkey:
Gives ~25% FPS boost on Aperture Desk Job map at `point_devshot_camera` camera by avoiding Dictionary usage in animations code.

Should decrease both CPU and RAM usage, but it's hard to measure actual RAM usage because of the GC.

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
@@ -38,7 +38,8 @@ namespace ValveResourceFormat.ResourceTypes.ModelAnimation
             }
         }
 
-        public static IEnumerable<Animation> FromData(IKeyValueCollection animationData, IKeyValueCollection decodeKey)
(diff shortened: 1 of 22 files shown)


ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/Animation.cs
```diff
@@ -90,38 +91,62 @@ public static IEnumerable<Animation> FromData(IKeyValueCollection animationData,
...
+                {
+                    "Position" => AnimationChannelAttribute.Position,
+                    "Angle" => AnimationChannelAttribute.Angle,
+                    "Scale" => AnimationChannelAttribute.Scale,
+                    _ => AnimationChannelAttribute.Unknown,
+                };
+
+                if (channelAttribute == AnimationChannelAttribute.Unknown)
+                {
+                    Console.WriteLine($"Unknown channel attribute '{localChannel.ChannelAttribute}' encountered with '{decoder}' decoder");
```
xPaw, line 113: This now spams `Unknown channel attribute 'data'`, that's what the `case "data"` ignore was for.

PR #501, opened 2022-12-30
Optimise InitializeTreeViewFromPackage

SamboyCoding:
I've made some changes to significantly reduce the time spent in `GUI.Types.Viewers.Package::Create`, specifically `TreeViewWithSearchResults::InitializeTreeViewFromPackage`. The way this code was previously set up caused thousands of ultimately unnecessary comparisons, which I've reduced (by setting the comparer before adding items) and I've also optimized the comparer to avoid type checks where possible - this does make use of `Unsafe`, however.

Time spent in the `InitializeTreeViewFromPackage` on my benchmark file, which was `a1_intro_world_2.vpk` in Half-Life Alyx, went from 75.2 seconds to 2.5 seconds, with the time spent in the tree view comparer now only 12ms, and total time to open the file went from 75.6 seconds to 2.7 seconds.


xPaw:
Merged in f0e5515ea394d6a2dec54283942c66920193f747

PR #502, opened 2022-12-30
Optimize texture loading

SamboyCoding:
Optimized the memory usage and memory copies when loading materials/textures by, where possible, using a direct reference to the memory-mapped file, instead of copying the relevant data up to 3 times per texture.

Tested on the same file as my other PR (HL:A `a1_intro_world_2.vpk`), all materials looked the same and:

- Final memory usage once everything had settled went from 11.1gb -> 10.5gb
- Time to render the first frame, from double-clicking the map file, went from 32 -> 24 seconds.

Material loading for non-LZMA-compressed textures is now zero-alloc, and just returns a pointer to the memory-mapped file. This did require enabling unsafe code, but I think it's a good tradeoff. For LZMA-compressed textures, a temporary buffer is rented from an array pool.

### DIFF
--- a/GUI/GUI.csproj
+++ b/GUI/GUI.csproj
@@ -11,6 +11,7 @@
     <IncludeAllContentForSelfExtract>true</IncludeAllContentForSelfExtract>
     <StartupObject>GUI.Program</StartupObject>
     <AssemblyName>VRF</AssemblyName>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


xPaw:
Merged in f0e5515ea394d6a2dec54283942c66920193f747

xPaw:
I have commited the ArrayPool usage in master, using it in binarykv3 decoding as well. I have tested your PR in release mode and I don't see huge loading time savings, so I am not feeling comfortable adding unsafe code (as we discussed on discord).

I might revisit the unsafe pointer from mmap into opengl directly, but it would be specifically contained in the texture loading code, without adding FastMemoryMappedFileStream.

Thanks for your time, but I'm not willing to merge this at this time.

PR #503, opened 2022-12-30
Flatten LINQ in CCompressedAnimQuaternion to a foreach

SamboyCoding:
Originally implemented in #502, split out to another PR as per request


xPaw:
Fixed in e28bf2bc6760e04050285c3c45954674a4940540 as well as for some other decoders.

PR #510, opened 2023-01-10
Scene model picker

### LINKED ISSUE #505: Add ability to select or pick models in sceneworlds
It would be nice if models in a map could be double-clicked so they could be opened separately.

image

kristiker:
To Do:
- [x] Fix `GL.ReadPixels` access violation
- [x] Debug view
- [x] Merge debug view with other modes
- [x] Only render to picker on double click.
- [x] Improve node search, maybe with a name cache.
- [x] Maybe outline the object on first click.
- [x] Test other scenes (model, particle)
- [ ] Fix 3D Skybox
- [x] Copy camera position when opening to separate tab
- [x] Select and identify model on scene using a separate frame buffer
- [x] Unbind events on dispose
- [ ] Better node search that supports changing ids, possibly incorporating octree node ids + queried with frustum
- [x] Double clicking in model viewer opens vmesh if available
- [x] Set meshgroups and animation same as in the map (e.g. tree model in dota)

Resolves #505

### DIFF
--- a/GUI/Types/Renderer/MeshBatchRenderer.cs
+++ b/GUI/Types/Renderer/MeshBatchRenderer.cs
@@ -15,8 +15,8 @@ namespace GUI.Types.Renderer
             public RenderableMesh Mesh;
             public DrawCall Call;
             public float DistanceFromCamera;
-            public int NodeId;
-            public int MeshId;
+            public uint NodeId;
+            public uint MeshId;
         }
 
         public static void Render(List<Request> requests, Scene.RenderContext context)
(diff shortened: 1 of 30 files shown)


GUI/Types/Renderer/MeshBatchRenderer.cs
```diff
@@ -48,7 +50,13 @@ private static void DrawBatch(IEnumerable<Request> drawCalls, Scene.RenderContex
             var cameraPosition = context.Camera.Location.ToOpenTK();
             var lightPosition = cameraPosition; // (context.LightPosition ?? context.Camera.Location).ToOpenTK();
 
-            foreach (var shaderGroup in drawCalls.GroupBy(a => a.Call.Shader))
+            var groupedDrawCalls = context.ReplacementShader switch
```
xPaw, line 53: This should be a ternary or a simple if. The switch is a bit silly.

PR #521, opened 2023-03-17
Add support for remap tables bigger than skeleton bone count

moofMonkey:
This reverts #520 and attempts to properly fix remap tables bigger than skeleton bone count.

I wouldn't consider it finished as eyes on the model from Dota 2 `models/items/witchdoctor/wd_ti8_immortal_bonkers/wd_ti8_immortal_bonkers.vmdl` still don't move, while eyelids do.


xPaw:
Closing this due to changes in 91f674c2d31cc595c713e9592cf6a38944b01bb6 which removed the Math.Min. Unsure what's up with that model since I can't even see eyeballs in the GUI.

Issue #525, opened 2023-03-24
Some normal maps have flipped green channels 


DarkBlade909:
### Describe the bug

Certain normal maps export with flipped green channels. Most of the normal maps in CS2, and at least one in Alyx

### What Valve game does this happen in

Half-Life:Alyx, CS2

### Reference file in a Source 2 game

In Half-Life: Alyx
materials/models/architecture/doors_1/door_1b_normal_tga_5e921617.vtex

In CS2:
weapons/models/usp_silencer/materials/usp_silencer_default_normal_tga_61d8e869.vtex

Most of the textures in Half-Life: Alyx don't to be flipped, so far I've only found the one above.
For example, this one exports fine:
models/characters/metrocop/materials/metro_head_normal_tga_b6a11f69.vtex

### Expected behavior

VRF Should detect whether the material is using a normal map green channel flip and export the texture accordingly

### What version of VRF are you using? On what platform?

v0.3.1 on Windows 10

xPaw:
Fixed by 816b3e3580279c8c39fb81c8a1bf7f9c2714823f and c41c73d87df11dff45f1ca4969d4764af9837ca4

PR #526, opened 2023-03-24
Export CS2 weapons with PBR

kristiker:
image


GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -77,10 +76,6 @@ vec3 calculateWorldNormal()
     vec3 tangentNormal = oct_to_float32x3(temp);
 #endif
 
-#if param_LegacySource1InvertNormals == 1
-    tangentNormal.y *= -1.0;
```
xPaw, line 81: Are we sure about this in HLA/SteamVR and stuff?

xPaw (changes requested):

GUI/Types/Renderer/Shaders/simple.frag
kristiker: The HLA door normal mentioned in #525 does appear to be in OpenGL format after removing invert on Decode (looking at the image). It matches with other normalmaps that don't have Source1 Invert. Since the VRF renderer is OpenGL it works nicely. S&box devs have mentioned that all their DirectX shaders have to do Y inversion.

Since inverted Y normals is an artist choice I don't think they have any logic in their shaders besides the having to flip on Vulkan.

xPaw: @lewa-j I doubt you remember, but do you know why we had to invert normals in the first place?

It was added in https://github.com/SteamDatabase/ValveResourceFormat/commit/10bf93bcf4fe39b761341484fbc37e6347516341 but it mentions hemioct, which is separate.

xPaw (approved):

PR #539, opened 2023-04-07
Vmap decompile

kristiker:
Work in progress, vmap_c / vpk map decompiler.

Goals:
- [x] World layers
- [x] Entity IO
- [ ] Export single WorldNode to vmap
- [ ] Export single EntityLump to vmap
- [ ] VMESH sceneobjects (old dota map, prop_static requires VMDL)
- [x] VPHYS to Mesh (tough, necessary)
  - [ ] My hammer mesh implementation is not stable enough. Normal/Tangent/UV data is not correct.
- [ ] VMDL to Mesh (tougher, not as necessary)
  - [ ] Seems to be necessary for aggregates.
  - [ ] Two alternatives here. Either to Mesh or
  - [x] To DMX. Can split up aggregate to .dmx + .vmdl files. Add vmdl as a prop_static.
- [ ] Binary encoding, right now it is text (kv2)
- [x] Reference DMX lib as package
  - [ ] Move VMAP format to the DMX lib
  - [ ] Match classnames etc with the sbox hammer api https://asset.party/api/Editor.MapDoc 
  (mostly CMapMesh -> MapMesh)

Things to note:
- Map meshes are exported as compiled models, and can be converted to editable mesh in editor. However:
  - It's all triangulated
  - It's all combined in in odd ways, based on the visibility algorithm that was used
  - Hammer does not like aggregate models found in Aperture and CS2
- Per-mesh physics and visibility exclusion data is lost
- Entities decompile fine, spare some unknown keys
- Selection set, prefab, instancing data is lost

Source asset: 
image

Vrf decompiled:
image

Skybox is in a phys group.


GUI/Types/Viewers/Resource.cs
```diff
@@ -449,22 +455,20 @@ static void VcsShaderResourceBridge(TabControl resTabs, SboxShader sboxShader)
...
-            {
-                foreach (TabPage tab2 in resTabs.TabPages)
-                {
-                    if (tab2.Text == "DATA")
-                    {
-                        resTabs.SelectTab(tab2);
+                    case ResourceType.Map:
+                    case ResourceType.World:
+                        var mapExtract = new MapExtract(resource, vrfGuiContext.FileLoader);
+                        var shouldFoucs = resource.ResourceType == ResourceType.Map;
```
xPaw, line 462: "Foucs" typo

kristiker:
Branch moved to the main repo. New Pull Request is at #568

PR #549, opened 2023-05-14
Texture viewer "View channels"

### LINKED ISSUE #194: No option to disable transparency in texture viewer
### Describe the bug
Often, textures will have an alpha channel. Newer versions of VRF will always render that as transparency with a black background, which for many textures results in difficulty viewing and is not consistent with how Source 2 displays them:
ValveResourceViewer_2020-03-04_19-18-09

### Reference file in a Valve game
Many `vtex_c` files in the Half-Life: Alyx SteamVR environment/asset pack VPKs have alpha channels that are not even encoding transparency information.
Example: `2006656803.vpk\materials\architecture\c17\c17_plaster_hs01a_mural_color_tga_387777a8.vtex_c`
In this example the alpha channel is just solid 0.5 and is clearly not supposed to represent transparency as this is an opaque material.

### Expected behavior
A checkbox to disable rendering transparency in the texture viewer would resolve this issue.

Additionally, an option to view the individual RGB or A channels (some textures have specific information in specific channels other than color/normal) would also be useful, Source 2's native Asset Viewer does this for textures:
image

### Version info
Most recent versions, including 0.1.0
In 0.0.6 (and possibly newer versions than that) transparency was not implemented and textures could be viewed more easily.

kristiker:
Closes #194 

| Opaque |
|:-------:|
|  |
| Transparent |
|  |

### DIFF
--- a/GUI/Forms/Texture.cs
+++ b/GUI/Forms/Texture.cs
@@ -1,17 +1,33 @@
+using System;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Windows.Forms;
+using GUI.Utils;
+using SkiaSharp;
+using Channels = ValveResourceFormat.CompiledShader.ChannelMapping;
 
 namespace GUI.Forms
 {
     public partial class Texture : UserControl
     {
         private string name;
+        private SKBitmap skBitmap;
+        private CancellationTokenSource cts;
+        private Task channelChangingTask;
 
         public Texture()
         {
             InitializeComponent();
+        }
+
+        public void SetImage(SKBitmap skBitmap, string name, int w, int h)
+        {
+            this.skBitmap = skBitmap;
+            viewChannelsTransparent.Enabled = skBitmap.AlphaType != SKAlphaType.Opaque;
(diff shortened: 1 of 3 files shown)


GUI/Forms/Texture.cs
```diff
@@ -1,26 +1,76 @@
...
+            {
+                Image image = null;
+                if (channels == Channels.RGBA || (skBitmap.AlphaType == SKAlphaType.Opaque && channels == Channels.RGB))
+                {
+                    image = skBitmap.ToBitmap();
+                }
+                else
+                {
+                    var pngBytes = ValveResourceFormat.IO.TextureExtract.ToPngImageChannels(skBitmap, channels);
+                    image = Image.FromStream(new MemoryStream(pngBytes));
```
kristiker, line 60: So this could be better. The Skia bitmap is kept as reference just for this `ToPngImageChannels` method, which also wastes time with the png encode.

kristiker: I tried doing a custom draw of the image with a color matrix, and it worked almost the same way, except some images would lose color data due to pre-multiplied alpha 😞

xPaw:
Perhaps the right click menu should be replaced with a toolbar (same with the save-as). Currently it takes too much clicks if you want to swap through channels.

PR #555, opened 2023-06-01
`env_sky` rendering

kristiker:
2D skybox rendering.
image
image

### DIFF
--- a/GUI/Types/Renderer/GLSceneViewer.cs
+++ b/GUI/Types/Renderer/GLSceneViewer.cs
@@ -146,6 +146,8 @@ namespace GUI.Types.Renderer
 
             GL.Enable(EnableCap.CullFace);
 
+            Scene.Sky?.Render(new Scene.RenderContext { Camera = e.Camera });
+
             if (ShowBaseGrid)
             {
                 baseGrid.Render(e.Camera, RenderPass.Both);
@@ -233,6 +235,12 @@ namespace GUI.Types.Renderer
         private void SetRenderMode(string renderMode)
         {
             ViewerControl.Camera?.Picker.SetRenderMode(renderMode);
+            if (Scene.Sky != null)
+            {
+                using var mat = GuiContext.LoadFileByAnyMeansNecessary(Scene.Sky?.Material.Material.Name + "_c");
+                Scene.Sky.Material = GuiContext.MaterialLoader.LoadMaterial(mat);
+            }
+
 
             foreach (var node in Scene.AllNodes)
             {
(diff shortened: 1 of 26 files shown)


GUI/Types/Renderer/MaterialLoader.cs
```diff
@@ -88,7 +88,14 @@ bool TryBindTexture(RenderMaterial mat, string name, string path)
             {
                 if (mat.Shader.GetUniformLocation(name) != -1)
                 {
-                    mat.Textures[name] = LoadTexture(path);
+                    if (name != "g_tSkyTexture")
```
xPaw, line 91: Is checking name going to work for cubemap entities too?

GUI/Types/Renderer/GLSceneViewer.cs
```diff
@@ -233,6 +235,12 @@ protected void SetEnabledLayers(HashSet<string> layers)
         private void SetRenderMode(string renderMode)
         {
             ViewerControl.Camera?.Picker.SetRenderMode(renderMode);
+            if (Scene.Sky != null)
```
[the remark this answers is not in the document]
kristiker: It is. This reloads the material, useful when iterating the shader.

GUI/Types/Renderer/MaterialLoader.cs
kristiker: I don't plan to hardcode this. I think I will end up adding (targetType, handle) as values to the Textures dictionary.

GUI/Types/Renderer/ShaderLoader.cs
```diff
@@ -172,7 +176,7 @@ private static string UpdateDefines(string source, IDictionary<string, bool> arg
                 var defaultValue = define.Groups["DefaultValue"];
                 var index = defaultValue.Index;
                 var length = defaultValue.Length;
-                var newValue = value ? "1" : "0";
+                var newValue = $"{value}";
```
xPaw, line 179: value.ToString()

GUI/Types/Renderer/ShaderLoader.cs
```diff
@@ -288,8 +294,8 @@ private uint CalculateShaderCacheHash(string shaderFileName, IDictionary<string,
 
             foreach (var key in parameters)
             {
-                shaderCacheHashString.AppendLine(key);
-                shaderCacheHashString.AppendLine(arguments[key] ? "t" : "f");
+                shaderCacheHashString.Append(key);
+                shaderCacheHashString.Append(Convert.ToBase64String(new[] { arguments[key] }));
```
xPaw, line 298: Isn't it a byte, just doing ToString should be enough on it?

kristiker: Murmurhash didn't like numbers, this fixed the collisions. Presumably that's why it used to have "t" and "f".

kristiker: Nevermind ToString works fine now 😄

xPaw (approved):

PR #559, opened 2023-06-16
Cubemap tweaks + gamma correction fixes

Purrspctiv:
- Added cubemap normalization, used in HLA (more info: page 30, https://ubm-twvideo01.s3.amazonaws.com/o1/vault/gdc2019/presentations/Hobson_Josh_The_Indirect_Lighting.pdf)
- Tweaked some things with envmaps that are used in HLA
- Fixed gamma correction so that it is only applied at the very end. This prevents incorrect multiplication of already gamma-corrected data with other gamma-corrected data.
-Added utils.glsl, for all the utility functions that came in useful in my own version of the renderer.

Normalization is currently not fully complete, as it won't cancel out the original lighting. I'm unsure how they calculate the values they use for this part. The rest of the code is fully implemented, though.


GUI/GUI.csproj
```diff
@@ -32,6 +32,9 @@
     <EmbeddedResource Include="AssetTypes\*" />
     <EmbeddedResource Include="Types\Renderer\Shaders\**" />
   </ItemGroup>
+  <ItemGroup>
```
xPaw, line 35: These csproj changes VS likes to do sometimes are fluff and not required.

GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -353,18 +351,20 @@ void main()
...
 
-    outputColor.rgb *= (irradiance * kD);
+    outputColor.rgb *= (irradiance);
     outputColor.rgb += Lo;
 
+    // Environment Map
     #if (S_SPECULAR == 1)
-        float layer = sqrt(roughness * roughness) * MAX_ENVMAP_LOD;
-        vec3 specular = F * pow(GetEnvironment(R, layer), power);
+        vec3 specular = F * GetEnvironment(N, V, roughness, F0, irradiance);
```
xPaw, line 361: `F` was removed

GUI/Types/Renderer/Shaders/common/environment.glsl
```diff
@@ -38,12 +38,63 @@ vec3 CubeMapBoxProjection(vec3 pos, vec3 R, vec3 mins, vec3 maxs, vec3 center)
...
+        // Reduce cancellation on glossier surfaces
+        float NormalizationClamp = max(1.0, rough * CubemapNormalizationParams.x + CubemapNormalizationParams.y);
+        return min(NormalizationTerm, NormalizationClamp);
+    #else
+        return 1.0;
+    #endif
+}
+
+// BRDF
+uniform sampler2D g_tBRDFLookup;
```
kristiker, line 72: Gonna have to handle the cases where this texture can't be found.

Issue #180, opened 2020-02-02
Allow selecting multiple files with shift to export multiple files at once with context menu


Wiethoofd:
In the file navigation (on the right side when opening a vpk) you are able able to select multiple items (shift and ctrl are allowed for selection).

However, when select multiple files in an overview and then right-click the context menu it offers the option to 'Export' and 'Copy file name'. This action is only applied to the first item (either file or folder) that was selected.

No 'Save as...' option is offered as on individual items either in overviews, where 'Export' in the menu on top is offering the same action, but in a wall of items if you've been browsing for a while. (see ~~#174~~ #149)

---

The main reason however for this issue is the ability to open a bunch of files in to new tabs by pressing *Enter* (or a right-click action to 'Open/View') after selecting multiple files in an overview.

In case a folder would be selected opening just that folder in a new tab (if not the only folder or root of the vpk) could be nice to have more detailed look at it's content by moving the overview to the left side and its content to the right side where it was previously.

tgrgds:
Can I have a crack at this?

It looks like since you can only multiple-select items in the `BetterListView`, we could just add a method like `ExportFile.ExtractFilesFromListViewNodes` to handle exporting anything from that view in particular?

If there's only one node selected we just go back to `ExportFile.ExtractFilesFromTreeNode` otherwise we have the user pick a path (like the folder exports) then go crazy >:)

i.e. in the Main Form's [`ExtractFiles`](https://github.com/SteamDatabase/ValveResourceFormat/blob/ad10461f8ec0dcd9ecd75afe272bb936b41cf953/GUI/MainForm.cs#L779) method

```diff
else if (owner.SourceControl is BetterListView listView)
{
-    foreach (ListViewItem selectedNode in listView.SelectedItems)
-    {
-        Debug.WriteLine(selectedNode);
-        ExportFile.ExtractFilesFromTreeNode((BetterTreeNode)selectedNode.Tag, listView.VrfGuiContext, decompile);
-    }

+    ExportFile.ExtractFilesFromListViewNodes(listView.SelectedItems, listView.VrfGuiContext, decompile);
}
```

Then in `ExtractFilesFromListViewNodes`
```c#
public static void ExtractFilesFromListViewNodes(BetterListView.SelectedListViewItemCollection items, VrfGuiContext vrfGuiContext, bool decompile)
{
    if (items.Count == 1)
    {
[...]

PR #572, opened 2023-07-02
Support for multiple-select export

tgrgds:
- Add `ExportFile.ExtractFilesFromListViewNodes`
- Handle non-folder nodes being passed into `ExtractProgressForm.QueueFiles`

As mentioned in comments, the potential issue with this is that the exported files will preserve the original tree instead of being placed directly in the folder the user asks for.

Alternatively `ExtractFileFromStream` could be used for nodes that are files, but I would've had to rip out the dialog popup inside it, or make it optional or something which is probably not as clean

Addresses #180/#531

### DIFF
--- a/GUI/Forms/ExtractProgressForm.cs
+++ b/GUI/Forms/ExtractProgressForm.cs
@@ -143,23 +143,36 @@ namespace GUI.Forms
 
         public void QueueFiles(BetterTreeNode root)
         {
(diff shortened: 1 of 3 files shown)


xPaw:
Thanks. There's certainly this that could be improved code wise (like figuring out how to handle all of this in less code), but I'll merge it.

PR #577, opened 2023-07-13
Various shader improvements

Purrspctiv:
- Added cloth shading (currently uses HLA code, which may have been updated in CS2)
- Added retroreflectivity support
- Refactor of how lighting is stored (now uses a "LightingTerms" struct)
- BRDF is now more accurate to S2 code
- New ShaderID rendermode. This could later be updated to automatically get values from gameinfo.gi, to account for different values for generic.vfx across games
- Some of the VR materials are now supported better
- Changed the fullbright rendermode to be more accurate. It's really not that much different than before
- Tint is now calculated more accurately
- Added texturing.glsl and moved some functions there
- Numerous tiny things that HLA had are now supported, like geometric roughness, centroid normal switch and alpha test AA. Some are noticeable, most aren't, but they're there now anyways.
- Many other tweaks and optimizations


GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -327,60 +400,87 @@ void main()
...
+    // Indirect Lighting
     #if (D_BAKED_LIGHTING_FROM_LIGHTMAP == 1) && (LightmapGameVersionNumber > 0)
         irradiance = texture(g_tIrradiance, vLightmapUVScaled).rgb;
         vec4 vAHDData = texture(g_tDirectionalIrradiance, vLightmapUVScaled);
-        const float DirectionalLightmapMinZ = 0.05;
-        irradiance *= mix(1.0, vAHDData.z, DirectionalLightmapMinZ);
-        occlusion *= vAHDData.w;
+
+        irradiance = ComputeLightmapShading(irradiance, vAHDData, normal);
+        //occlusion *= vAHDData.w; // vAHDData.w is actually unused
```
kristiker, line 417: You sure? There's baked AO in there (depends on map)

GUI/Types/Renderer/Shaders/simple.vert
```diff
@@ -86,19 +106,17 @@ void main()
...
 #endif
 
-    vTexCoordOut = vTEXCOORD;
+	vTexCoordOut = vTEXCOORD * g_vTexCoordScale.xy + g_vTexCoordOffset.xy + (g_vTexCoordScrollSpeed.xy * g_flTime);
 
 #if D_BAKED_LIGHTING_FROM_LIGHTMAP == 1
     vLightmapUVScaled = vec3(vLightmapUV * g_vLightmapUvScale, 0);
 #elif D_BAKED_LIGHTING_FROM_VERTEX_STREAM == 1
+    //vec3 Light = vPerVertexLighting.rgb * 6.0 * vPerVertexLighting.a;
+    //vPerVertexLightingOut = pow2(Light);
```
kristiker, line 120: RGBM right?

GUI/Types/Renderer/Shaders/simple.frag
```diff
@@ -327,60 +418,91 @@ void main()
...
 
+    vec3 combinedLighting = diffuseColor * diffuseLighting + specularLighting;
+
+    outputColor.rgb = combinedLighting;
+
+#endif
+#if F_DISABLE_TONE_MAPPING == 0
     outputColor.rgb = pow(outputColor.rgb, vec3(invGamma));
     //outputColor.rgb = SRGBtoLinear(outputColor.rgb);
 #endif
```
kristiker, line 471: Shouldn't it be inside the #endif? I imagine unlit materials don't not need any correction.

Purrspctiv: Funnily enough, it's actually the reverse: vr_complex only exposes this parameter if the material is unlit.

PR #599, opened 2023-08-11
feat: add devcontainer and vscode config

pjoe:
This adds devcontainer and vscode configuration, which make it possible to just clone project, open in vscode devcontainer and cross compile GUI project (devcontainer is linux).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,27 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+	"name": "C# (.NET)",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:0-7.0"
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [5000, 5001],
+	// "portsAttributes": {
+	//		"5001": {
+	//			"protocol": "https"
+	//		}
+	// }
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "dotnet restore",
+
+	// Configure tool-specific properties.
+	// "customizations": {},
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            // Use IntelliSense to find out which attributes exist for C# debugging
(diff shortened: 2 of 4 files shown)


xPaw:
Does this *actually work*?

Lifeismana:
On what platform did you try that?
On windows it doesn't work for me (getting a compilation error)

pjoe:
This is on windows (with docker in wsl), you need to make sure you (re)open the project in the devcontainer and let it build the container and install required dotnet tooling (which it should do automatically - but may take a bit first time). Then Ctrl-Shift-B to build the GUI project. I'll just do a test to see if it works in Codespaces too

pjoe:
Starting in Codespaces:
Screenshot from 2023-08-11 19-34-09

Compiled inside Codespaces:
Screenshot from 2023-08-11 19-39-49

Lifeismana:
Found the solution to following error which was to `dotnet clean` outside of the container (it probably didn't like that there was already a build outside of it)
```
/usr/share/dotnet/sdk/7.0.306/Microsoft.Common.CurrentVersion.targets(4884,5): error MSB3374: The last access/last write time on file "/workspaces/ValveResourceFormat/Decompiler/obj/Debug/Decompiler.csproj.CopyComplete" cannot be set. Access to the path '/workspaces/ValveResourceFormat/Decompiler/obj/Debug/Decompiler.csproj.CopyComplete' is denied. [/workspaces/ValveResourceFormat/Decompiler/Decompiler.csproj]
```

But now i'm getting the following error when running VRF 
```
A fatal error was encountered. The library 'libhostpolicy.so' required to execute the application was not found in '/workspaces/ValveResourceFormat/GUI/bin/Debug/win-x64/'
```

pjoe:
Ok, how exactly are you running it? Which project? I was actually only building inside the devcontainer and then running the resulting VRF.exe gui app from the windows host. dotnet gave me warnings that the GUI project would only run on windows so it had to be cross compiled, however I'd expect the CLI to work fine on linux, but haven't checked that ... yet

pjoe:
Running tests passes for me in Codespaces:

image

Lifeismana:
Ah, okay, you're just using the container to compile VRF, that's kind of a weird usage, why not just install .net's sdk ?
i was trying to run vrf in the container, that's what's not working, you might want to remove the launch.json if that's not usable.

I kinda don't see the usefulness of this in the repo if you can't run it inside the container

pjoe:
Right, the GUI app won't run in container, even less in Codespaces, decompiler does though.

Will change launch.json to run the decompiler.

For me point of running in devcontainer, is that I don't need to have any SDK stuff installed on the host (I have enough of those already) - relying on the container setup I can be sure to get all need tools in compatible version just by "run inside devcontainer".

Lifeismana:
don't think having decompiler in launch.json make sense either since it doesn't have an "interactive aspect"

i'll stick to my "it doesn't really make sense to have this merged" if doesn't allow the gui to be used (one thing i didn't try but saw mentioned was to build via the publish command `dotnet publish` and to target 'linux-x64', this could make it have every lib it needs)

(FYI: if this doesn't get merged and you want to hide this from git you can add it to the exclude file in `.git/info/exclude`)

pjoe:
No worries. I have in general found it helpful to add devcontainer support to projects for others to get started on it quickly, but you are right that the benefit in this case is minimal. I'll close this PR

PR #609, opened 2023-08-27
Gradient/Cubemap Fog support

Purrspctiv:
This adds support for gradient fog and cubemap fog, which are used in Half-Life Alyx and later.

The remaining fog types include spherical vignettes (which won't be implemented as they are meant for restricting visibility during cinematics and would destroy visibility) and volumetric fog. Volumetric fog would be difficult to implement but would give a massive payoff. It isn't used in CS2, though- it is only used in the VR renderer.

This improves two other things as a result of the fixes required to implement this:
- Reflections in Skybox maps are now fixed, which is especially noticeable in de_vertigo, as all buildings previously had broken reflections. This is because the shaders did not take into account the skybox scale when calculating g_vCameraPosition.
- Skybox textures are now gamma-corrected

Known issues:
- Slight technical discrepencies between the game and VRF values for the cubemap fog log2mip value. This is because I'm not quite sure how to actually calculate it. Won't be very noticeable.

### DIFF
--- a/GUI/Types/Renderer/Camera.cs
+++ b/GUI/Types/Renderer/Camera.cs
@@ -83,14 +83,14 @@ namespace GUI.Types.Renderer
             return new Vector3(MathF.Cos(Yaw - OpenTK.MathHelper.PiOver2), MathF.Sin(Yaw - OpenTK.MathHelper.PiOver2), 0);
         }
 
-        public UniformBuffers.ViewConstants SetViewConstants(UniformBuffers.ViewConstants viewConstants)
+        public UniformBuffers.ViewConstants SetViewConstants(UniformBuffers.ViewConstants viewConstants, float mapScale)
         {
             return viewConstants with
             {
                 WorldToProjection = ProjectionMatrix,
                 WorldToView = CameraViewMatrix,
                 ViewToProjection = ViewProjectionMatrix,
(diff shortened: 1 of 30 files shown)


GUI/Types/Renderer/Camera.cs
```diff
@@ -83,14 +83,14 @@ private Vector3 GetRightVector()
             return new Vector3(MathF.Cos(Yaw - OpenTK.MathHelper.PiOver2), MathF.Sin(Yaw - OpenTK.MathHelper.PiOver2), 0);
         }
 
-        public UniformBuffers.ViewConstants SetViewConstants(UniformBuffers.ViewConstants viewConstants)
+        public UniformBuffers.ViewConstants SetViewConstants(UniformBuffers.ViewConstants viewConstants, float mapScale)
```
xPaw, line 86: Camera instance already has a `Scale` property

GUI/Types/Renderer/RenderMaterial.cs
```diff
@@ -52,7 +58,7 @@ public RenderMaterial(Material material, IKeyValueCollection insg, ShaderLoader
                 isAdditiveBlend = blendMode == 4;
             }
         }
-
+
```
xPaw, line 61: whitespace

PR #633, opened 2023-10-03
feat: decompiler export gltf with textures

pjoe:
Support exporting gltf with all mats/texs when using decompiler CLI.

This is done by "combining" `AdvancedGuiFileLoader` and `BasicVpkFileLoader`

### DIFF
--- a/GUI/Utils/AdvancedGuiFileLoader.cs
+++ b/GUI/Utils/AdvancedGuiFileLoader.cs
@@ -184,7 +184,7 @@ namespace GUI.Utils
             return collection;
         }
 
-        public ShaderCollection LoadShader(string shaderName)
+        public new ShaderCollection LoadShader(string shaderName)
         {
             lock (shaderCacheLock)
             {


Decompiler/Decompiler.cs
```diff
@@ -848,7 +848,7 @@ private void DumpVPK(string parentPath, Package package, string type, Dictionary
                 return;
             }
 
-            var fileLoader = new BasicVpkFileLoader(package);
+            var fileLoader = new GameFileLoader(package);
```
xPaw, line 851: I think this needs to pass in filename too?

ValveResourceFormat/IO/GameFileLoader.cs
```diff
@@ -0,0 +1,482 @@
...
+                resource.Read(stream);
+                return resource;
+            }
+
+            return null;
+        }
+
+        public ShaderCollection LoadShader(string shaderName) => null;
+
+        public static Stream GetPackageEntryStream(Package package, PackageEntry entry)
```
xPaw, line 436: Let's not move this into lib code. This should be done in https://github.com/ValveResourceFormat/ValvePak/issues/10

pjoe: oh, where did that come from, my bad will fix

pjoe: hmmm seems like it is needed by IFileLoader?

PR #636, opened 2023-10-05
Add converted MurmurHash Value on the Readme to make it copy paste ready for the online tool

karl-police:
http://murmurhash.shorelabs.com/

This one, since it's useful for Source Filmmaker materialOverrides, even though you can type the string as well. But if you do not know the string and want to figure it out, then it's useful.


xPaw:
I don't think this is needed.

karl-police:
> I don't think this is needed.

If you say so 🤔
What do you do if you need this number to figure out the original string of a Source 2 Filmmaker key value thing?

PR #662, opened 2023-11-11
Export animations when decompiling vmdls

krotter0:
Frame class doesn't load morphs, so only bone channels are exported.

### DIFF
--- a/ValveResourceFormat/IO/ContentFormats/DmxModel.cs
+++ b/ValveResourceFormat/IO/ContentFormats/DmxModel.cs
@@ -1,6 +1,7 @@
 using DMElement = Datamodel.Element;
 using Datamodel.Format;
 using System.Numerics;
+using System;
 
 namespace ValveResourceFormat.IO.ContentFormats.DmxModel;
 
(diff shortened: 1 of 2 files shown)


xPaw:
Gltf does handle morphs, did you look at that? Is there something extra that is not yet handled?

krotter0:
I don't see a way to get morph animation data through the Animation class, Frame class only contains data for bones too. I can see morphs in exported gltfs, but they don't seem to be animated either (unless I'm missing something obvious?)

xPaw:
Morph exists separately (LoadExternalMorphData), I'm not entirely sure if there is any morph data that can exists on animations/frames.

krotter0:
They can, one example would be the @catwalk_nag animations of models/creatures/vortigaunt/uncombined_vortigaunt.vmdl_c in HLA.

I've managed to implement MorphChannels in another branch, but I feel like there'd need to be some refactoring on some of the animation classes for a clean implementation. I won't make a pr for it, but it's here if you'd want to take a look at how it works: https://github.com/krotter0/ValveResourceFormat/tree/animation_export_flexes

xPaw:
There might be more success in refactoring that by adding support for animating morphs in the model viewer, that should propagate all the necessary changes that then can be used by the exporter.

xPaw:
I merged this, do feel free to figure out morphs. Jump on discord if you want to discuss it

PR #670, opened 2023-11-20
Model morphs implementation

krotter0:
Morph implementation for model viewer and model extract, including refactoring of anim segment decoder classes.


ValveResourceFormat/Resource/ResourceTypes/Morph.cs
```diff
@@ -131,8 +129,83 @@ public void LoadFlexData(IFileLoader fileLoader)
...
+        {
+            if (obj is not IKeyValueCollection kv)
+            {
+                throw new ArgumentException("Parameter is not IKeyValueCollection");
             }
+
+            var name = kv.GetStringProperty("m_szName");
+            var min = kv.GetFloatProperty("min");
+            var max = kv.GetFloatProperty("max");
+            //There's also a m_szType field, but it seems to be always "default"
```
xPaw, line 172: Throw if it's not?

GUI/Types/Renderer/ModelSceneNode.cs
```diff
@@ -116,6 +131,19 @@ public override void Update(Scene.UpdateContext context)
                 }
 
                 LocalBoundingBox = newBoundingBox;
+
+                var datas = AnimationController.GetDatas();
```
xPaw, line 135: this doesn't need to be in try/finally as it doesn't access matrices.

that being said, calling render from update isn't a very good idea.

GUI/Types/Renderer/Shaders/common/morph.glsl
```diff
@@ -0,0 +1,17 @@
...
+    int vertexId = gl_VertexID + morphVertexIdOffset;
+    return vec2(
+        1.5/2048 + mod(vertexId, morphCompositeTextureSize.x * 1.0) / 2048.0,
+        1 - (1.5/2048 + floor(vertexId / morphCompositeTextureSize.x * 1.0) / 2048.0)
+    );
+}
+
+vec3 getMorphOffset()
+{
+    return texture(morphCompositeTexture, getMorphUV()).xyz;
```
kristiker, line 16: Can we do #define D_MORPH 0
and 
#if (D_MORPH == 1) ..  else vec3(0, 0, 0)

These defines are passed as shader arguments at rendermesh creation time.

GUI/Types/Renderer/ModelSceneNode.cs
krotter0: What would be a good place for rendering them?

xPaw: Realistically it should be in Render(), but batch renderer doesn't call that for models as the comment implies, maybe that can be changed.

GUI/Types/Renderer/ModelSceneNode.cs
```diff
@@ -83,10 +83,25 @@ public ModelSceneNode(Scene scene, Model model, string skin = null, bool optimiz
...
+        {
+            foreach (var renderableMesh in activeMeshRenderers)
+            {
+                if (renderableMesh.MorphComposite == null || renderableMesh.FlexStateManager == null)
+                {
+                    continue;
+                }
+
+                renderableMesh.MorphComposite.SetMorphsFromFlexes(renderableMesh.FlexStateManager);
+                renderableMesh.MorphComposite?.Render();
```
kristiker, line 96: Null checks are incorrect.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/AnimationFrameCache.cs
```diff
@@ -35,6 +41,10 @@ public void Clear()
         /// <param name="time">The time to get the frame for.</param>
         public Frame GetInterpolatedFrame(Animation anim, float time)
         {
+            if (lastAnimation == anim && Math.Abs(lastTime - time) < 0.0001)
```
xPaw, line 44: What is this for?

GUI/Types/Renderer/ModelSceneNode.cs
```diff
@@ -271,13 +301,15 @@ public void SetAnimation(Animation activeAnimation)
                 foreach (var renderer in meshRenderers)
                 {
                     renderer.SetAnimationTexture(animationTexture, bonesCount);
+                    renderer.FlexStateManager?.ResetControllers();
```
xPaw, line 304: Should this be called inside of SetAnimationTexture instead?

GUI/Types/Renderer/GLSceneViewer.cs
```diff
@@ -257,6 +257,7 @@ void UpdateSceneBuffers(Scene scene, Camera camera)
 
             GL.DepthRange(0, 0.95);
 
+            GL.Viewport(0, 0, GLControl.Width, GLControl.Height);
```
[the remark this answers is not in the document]
krotter0: Morph composite rendering changes the viewport size, this resets it.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/AnimationFrameCache.cs
krotter0: ModelSceneNode.Update calls AnimationController.GetAnimationMatrices and AnimationController.GetDatas, which both call this function. I thought it would make sense that it returns a cached frame if it's called for the same frame, given that the class already stores the interpolated frame without doing much with it, and it being called AnimationFrameCache.

GUI/Types/Renderer/ModelSceneNode.cs
krotter0: I guess FlexStateManager storing a MorphComposite instance could make sense.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/AnimationFrameCache.cs
xPaw: I don't like this kind of code, it means we don't have full grasp of the flow here, ideally getting a frame during update would be called only once.

krotter0: I've pushed a different implementation just now

GUI/Types/Renderer/FlexStateManager.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+using ValveResourceFormat.ResourceTypes;
+using ValveResourceFormat.ResourceTypes.ModelFlex;
+
+namespace GUI.Types.Renderer
+{
+    class FlexStateManager
+    {
+        public Morph Morph { get; }
+        private Dictionary<string, int> controllerNameToId = new();
```
xPaw, line 13: Can we make Datas id based so the name lookups are not necessary?

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/AnimationFrameCache.cs
```diff
@@ -58,6 +59,15 @@ public Frame GetInterpolatedFrame(Animation anim, float time)
                 InterpolatedFrame.Bones[i].Scale = frame1Bone.Scale + (frame2Bone.Scale - frame1Bone.Scale) * t;
             }
 
+            var dataNames = frame1.Datas.Keys.Union(frame2.Datas.Keys);
```
xPaw, line 62: note: need to verify that this PR doesn't add some allocations per-frame

GUI/Types/Renderer/MorphComposite.cs
```diff
@@ -0,0 +1,316 @@
...
+
+            GL.EnableVertexAttribArray(positionWeightsLocation);
+            GL.EnableVertexAttribArray(texCoordsLocation);
+            GL.EnableVertexAttribArray(offsetsLocation);
+            GL.EnableVertexAttribArray(rangesLocation);
+
+            GL.BindVertexArray(0);
+        }
+
+        private void FillVerticies()
```
kristiker, line 163:
```suggestion
        private void FillVertices()
```

ValveResourceFormat/Resource/ResourceTypes/ModelFlex/FlexRule.cs
```diff
@@ -0,0 +1,38 @@
...
+                throw new ArgumentException("Flex ops array cannot be empty");
+            }
+
+            FlexID = flexID;
+            FlexOps = flexOps;
+        }
+
+        public float Evaluate(float[] flexControllerValues)
+        {
+            var context = new FlexRuleContext(flexControllerValues);
```
xPaw, line 28: This allocates every frame

kristiker (approved):

PR #676, opened 2023-12-08
vmdl extract: recreate skeleton in vmdl file

kristiker:
Recreated skeleton structure in vmdl file.

image

Decompiled animation poses the skeleton correctly, however the mesh does not follow it.

image


kristiker:
Looks like bone structure is mostly ok, and the bad visuals come from bad indices in dmx render mesh.

xPaw (approved):

PR #683, opened 2023-12-14
Extract animation root motion with .vmdl decompile

### LINKED ISSUE #665: Animation Export - Extracted motion

Animations with animated rootmotion that was extracted when the model was compiled (i.e. walk animations) do not have the motion represented in the model viewer or exported when decompiled using .vmdl/.dmx animation export.

The extracted motion should both be represented in the model viewer (treadmill effect with the grid moving to represent motion) and the motion should be included in the exported .dmx animation.

If the extracted motion animations can be identified S2V could warn the user when exporting, provide a list of these specific animations in the console/message and prompt the user whether they want to continue exporting these animations in particular

krotter0:
Fixes #665

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/MotionFlag.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/MotionFlag.cs
@@ -1,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ValveResourceFormat.ResourceTypes.ModelAnimation
+{
+    [Flags]
+    public enum MotionFlag
+    {
+        TX = 64,
+        TY = 128,
+        TZ = 256,
+        RZ = 2048,
+        Linear = 4096,
+    }
+}
--- a/ValveResourceFormat/IO/ModelExtract.cs
+++ b/ValveResourceFormat/IO/ModelExtract.cs
@@ -828,7 +828,30 @@ public class ModelExtract
             anim.DecodeFrame(frame);
             frames[i] = frame;
         }
+        //Root motion
+        var rootPositionChannel = BuildDmeChannel<Vector3>($"_p", skeleton.Transform, "position", out var rootPositionLog);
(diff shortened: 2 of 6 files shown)


xPaw:
So is this movement something we can show/implement in our model viewer? If so, can you make a separate issue for that please.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/MotionFlag.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ValveResourceFormat.ResourceTypes.ModelAnimation
+{
+    [Flags]
+    public enum MotionFlag
```
xPaw, line 10:
```suggestion
    public enum EMotionFlag
```

And how did you come with up these value names? I see there's missing values.

krotter0:
> So is this movement something we can show/implement in our model viewer? If so, can you make a separate issue for that please.

I have another branch that modifies the model's transforms using the delta information from the movement data. There's a few more things to figure out there though, but I think this part of it was already very useful for anim exports, so I've made this PR.

ValveResourceFormat/Resource/ResourceTypes/ModelAnimation/MotionFlag.cs
krotter0: These were the values that I've managed to reproduce using ModelDoc through the ExtractMotion node. The missing values between TZ and RZ are most likely RX and RY, but I didn't want to add guesses.

krotter0: This would be the first prefixed enum name in the solution this way, should I go ahead with renaming it?

xPaw: Whoops, I had a different project in mind.

Let's put it into Resource/Enums as `ModelAnimationMotionFlags` instead.

xPaw (approved):

PR #697, opened 2024-01-03
GPU texture decoding

kristiker:


kristiker:
Just found out about [glGetTexImage](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetTexImage.xhtml). Can be used instead of going through render target when the data is just passed-through i.e when there's no channel operations, or hemioctahedral normal map conversion.

GUI/Types/Renderer/MaterialLoader.cs
```diff
@@ -260,7 +266,6 @@ public RenderTexture LoadTexture(Resource textureResource, bool srgbRead = false
 
             // Dispose texture otherwise we run out of memory
             // TODO: This might conflict when opening multiple files due to shit caching
-            textureResource.Dispose();
```
kristiker, line 263: .

PR #724, opened 2024-02-14
Fixed rendering clipping brushes on some levels

mircokroon:
Rendering of clipping brushes seems to be broken when there's capsule shaped clipping brushes in the level. 

Current version:
image

Issue fixed:
image

Seems to be caused by these:
image

As far as I can tell from spending far too long looking at the code to find the issue, the problem is adding an incorrect number of vertex indices to the index buffer. Since it's expecting 3 per triangle and capsules add 8. Fixed it by adding a duplicate of one of the indices, not sure if there's a better way. The rendering of capsules looks a bit off to me but I'm not sure if this was already the case before or if it's caused by this fix.

I think the same issue is also present in circles, although since `AddCircle` is only used in `AddSphere` which calls it 3 times it shouldn't cause any issues with rendering the rest of the map.

### DIFF
--- a/GUI/Types/Renderer/PhysSceneNode.cs
+++ b/GUI/Types/Renderer/PhysSceneNode.cs
@@ -308,6 +308,7 @@ namespace GUI.Types.Renderer
 
                 inds.Add(baseVertex + i * 2);
                 inds.Add(baseVertex + i * 2 + 1);
+                inds.Add(baseVertex + i * 2);
             }
         }
 


kristiker (approved):

xPaw:
The problem is that we used to render physics as lines but then switched to triangles. And spheres are not triangulated, as can be seen from your last screenshot.

I'll merge this so it doesn't break everything else, but it would be nice to update spheres to be triangulated.

PR #729, opened 2024-02-17
Capsule / Sphere mesh rendering for clipping

mircokroon:
Triangulated capsules for clipping (related to #724)

image

(offset to show the full shape)
image

Made them yellow, not sure if the colors have any meaning so just went with that. I also changed the top/bottom of the capsule to hemispheres instead of full spheres, think that looks better with the transparent color than having it be a full sphere.


xPaw:
The colors are just to indicate different shapes (meshes, spheres, etc), so there is no inherent meaning.

xPaw:
Merged as 89693c844e550d0ad08174d6da2135ed027f62d2, I fixed the formatting.

PR #732, opened 2024-02-18
Read and decompile model attachments

krotter0:


ValveResourceFormat/IO/ModelExtract.cs
```diff
@@ -268,6 +270,42 @@ Lazy<KVObject> MakeLazyList(string className)
 
                 AddItem(renderMeshList.Value, renderMeshFile);
             }
+
+            var mesh = RenderMeshesToExtract.First();
```
kristiker, line 274: Any idea if the attachments ever change through meshes on the same model? Found it odd that they hold that data in MDAT when it doesn't change.

krotter0: I don't think they can change, I've always thought that they're just a part of the overall model. It's kind of implied by ModelDoc's UI too.

ValveResourceFormat/Resource/ResourceTypes/Mesh.cs
```diff
@@ -28,10 +30,21 @@ public VBIB VBIB
 
         private VBIB cachedVBIB { get; set; }
 
+        public Attachments Attachments { get; private set; }
+
```
kristiker, line 34: This might be ok, but since the data is stored with key="name" value=attachment, I thought maybe it can be a `Dictionary<string, Attachment>`. Would have to verify if models can ever have attachments with same name (by checking existing models with decompiler -stats , or experimenting in modeldoc).

ValveResourceFormat/Resource/ResourceTypes/ModelData/Attachments/Attachment.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var influenceNames = valueData.GetArray<string>("m_influenceNames");
+            var influenceRotations = valueData.GetArray("m_vInfluenceRotations", v =>
+            {
+                var x = (float)v.GetDoubleProperty("0");
+                var y = (float)v.GetDoubleProperty("1");
+                var z = (float)v.GetDoubleProperty("2");
+                var w = (float)v.GetDoubleProperty("3");
+                return new Quaternion(x, y, z, w);
+            });
```
kristiker, line 45: try
```suggestion
            var influenceRotations = props.GetArray("m_vInfluenceRotations").Select(v => v.ToQuaternion()).ToArray();
```

ValveResourceFormat/Resource/ResourceTypes/ModelData/Attachments/Attachment.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    public class Attachment : IEnumerable<Attachment.Influence>
+    {
+        public struct Influence
+        {
+            public string Name { get; init; }
+            public Vector3 Offset { get; init; }
+            public Quaternion Rotation { get; init; }
+            public double Weight { get; init; }
+        }
```
kristiker, line 15:
```suggestion
        public record struct Influence(string Name, Quaternion Rotation, Vector3 Offset, float Weight, bool RootTransform);
```
Weight is a float in engine. Kv3 storing as double is misleading. Also is RootTransform some legacy parameter that you didn't want to keep?

ValveResourceFormat/Resource/ResourceTypes/ModelData/Attachments/Attachment.cs
```diff
@@ -0,0 +1,77 @@
...
+        public bool IgnoreRotation { get; init; }
+
+        private readonly Influence[] influences;
+        public Influence this[int i]
+        {
+            get
+            {
+                return influences[i];
+            }
+        }
```
kristiker, line 27: If @xPaw is ok with inline arrays. Source2 CAttachment has fields of fixed size 3. Kv3 also always stores 3 influences. 
```suggestion
        [System.Runtime.CompilerServices.InlineArray(3)]
        public struct Influence3
        {
            public Influence Influence;
        }

        public Influence3 Influences { get; init; }
```

That lets you index into attachment.Influences with bracket notation [] as well.

ValveResourceFormat/Resource/ResourceTypes/ModelData/Attachments/Attachment.cs
krotter0: I'm not sure what the purpose of RootTransform is. It's set to true if the influence is the root transform, but in those cases, the influence name is also an empty string. It was missing from old .vmesh files.

kristiker: Okay then, if it's missing from old meshes and we make no use of for exporting/rendering then it can be ignored at this point.

ValveResourceFormat/Resource/ResourceTypes/ModelData/Attachments/Attachment.cs
krotter0: This can't be indexed if it's a property, and if it's not a property it's throwing 'CA1051 Do not declare visible instance fields'

ValveResourceFormat/Resource/ResourceTypes/Mesh.cs
krotter0: I've replaced this with a dictionary, ModelDoc doesn't allow naming two attachments the same either. The attachments class didn't make much sense anyways after looking at it.

kristiker (approved):

PR #744, opened 2024-03-09
Fix Angle Shear Vector for light_barn entity

Ansimist:
Before: 
before
After:
after

### DIFF
--- a/ValveResourceFormat/IO/ModelExtract.Mesh.cs
+++ b/ValveResourceFormat/IO/ModelExtract.Mesh.cs
@@ -325,7 +325,7 @@ partial class ModelExtract
 
                 var material = drawCall.GetProperty<string>("m_material") ?? drawCall.GetProperty<string>("m_pMaterial");
 
-                if (material != null && materialInputSignature.Elements.Length == 0 && materialInputSignatures != null)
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/IO/ModelExtract.Mesh.cs
```diff
@@ -325,7 +325,7 @@ private static void FillDatamodelVertexData(VBIB.OnDiskBufferData vertexBuffer,
 
                 var material = drawCall.GetProperty<string>("m_material") ?? drawCall.GetProperty<string>("m_pMaterial");
 
-                if (material != null && materialInputSignature.Elements.Length == 0 && materialInputSignatures != null)
+                if (material != null && materialInputSignatures != null && (materialInputSignature.Elements == null || materialInputSignature.Elements.Length == 0))
```
kristiker, line 328: Elements itself can't be null I think. Only zero length.

Ansimist: I don't know why but, I was getting an error at this point and it was breaking all the models

 after fixing it, I got this:

kristiker: Okay its likely because struct `default` doesn't initialize the properties.

kristiker (approved):

PR #747, opened 2024-03-11
Remove [PR#] prefix for Entity values

Ansimist:
Makes entity names and inputs/outputs visually cleaner and fixes some problems associated with this [PR#] prefix in the map editor

### DIFF
--- a/ValveResourceFormat/IO/MapExtract.cs
+++ b/ValveResourceFormat/IO/MapExtract.cs
@@ -851,7 +851,7 @@ public sealed class MapExtract
             }
 
             var key = property.Name;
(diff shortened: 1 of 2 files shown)


ValveResourceFormat/IO/MapExtract.cs
```diff
@@ -944,6 +944,8 @@ private static Vector3 GetVector3Property(EntityLump.EntityProperty property)
     static string StringBool(bool value)
         => value ? "1" : "0";
 
+    private static string RemovePrefix(string value) => value.Replace("[PR#]", "");
```
xPaw, line 947: StartsWith + Substring would be safer.

PR #760, opened 2024-04-08
.vfe implementation

### LINKED ISSUE #142: Add support for VFE files
VFE files seem to define face pose data - examples of decompiled VFE files can be found within CS:GO, inside the "expressions" directory.

krotter0:
Fixes #142


GUI/Types/Exporter/ExportFile.cs
```diff
@@ -117,6 +118,36 @@ public static void ExtractFileFromStream(string fileName, Stream stream, VrfGuiC
                     exportData.VrfGuiContext.Dispose();
                 }
             }
+            else if (decompile && fileName.EndsWith(".vfe", StringComparison.OrdinalIgnoreCase))
```
xPaw, line 121: Can we avoid special code for this here?

ValveResourceFormat/FaceExpressionData/FaceExpressionData.cs
```diff
@@ -0,0 +1,242 @@
...
+                if (Settings.TryGetValue(key, out var weight))
+                {
+                    return weight;
+                }
+
+                return default;
+            }
+        }
+
+        public const uint MAGIC = 0x00564645; //EFV\0
```
xPaw, line 55: Add to readme pls

ValveResourceFormat/FaceExpressionData/FaceExpressionData.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            var startPosition = reader.BaseStream.Position;
+
+            var magic = reader.ReadInt32();
+            if (magic != MAGIC)
+            {
+                throw new UnexpectedMagicException("Magic is not EFV", magic, nameof(magic));
+            }
+
+            Version = reader.ReadInt32();
```
xPaw, line 89: If this is static, throw on unexpected version

xPaw:
Add a test file, just checking that a full exported text file matches is enough.

ValveResourceFormat/FaceExpressionData/FaceExpressionData.cs
```diff
@@ -0,0 +1,242 @@
+using System.IO;
+using System.Text;
+using ValveResourceFormat.Utils;
+
+namespace ValveResourceFormat.FaceExpressionData
+{
+    public partial class FaceExpressionData
```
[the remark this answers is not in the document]
xPaw: I was going by the S1 code

kristiker (commented):
Looks ok. Maybe the `vfe.ToTextStream` is weird.

Now I realize S2V doesn't export assetinfo and some other special files like the decompiler does. They would also swap the stream here I guess.

xPaw:
> If it did, those files would also have to swap streams I guess.

We need to implement some shared/common way to export these.

GUI/Types/Exporter/ExportFile.cs
```diff
@@ -119,6 +120,16 @@ public static void ExtractFileFromStream(string fileName, Stream stream, VrfGuiC
             }
             else
             {
+                if (decompile && fileName.EndsWith(".vfe", StringComparison.OrdinalIgnoreCase))
```
xPaw, line 123: Maybe move this to `FileExtract` so that cli will also be able to use this with minimal changes (and then we can pull specific code from cli into this class as well so gui can export non resource files)

ValveResourceFormat/FaceExpressionData/FaceExpressionData.cs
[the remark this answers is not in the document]
xPaw: In S1, they have a CFlexSceneFileManager, which reads  "expressions/%s.vfe" files into CFlexSceneFile which contains the buffer (and then it's casted to flexsettinghdr_t etc)

```c
//-----------------------------------------------------------------------------
// Purpose:  A .vfe referenced by a scene during .vcd playback
//-----------------------------------------------------------------------------
class CFlexSceneFile
```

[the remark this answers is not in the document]
krotter0: Oh, I've missed that

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -225,6 +225,34 @@ public static ContentFile Extract(Resource resource, IFileLoader fileLoader)
             return contentFile;
         }
 
+        /// <summary>
+        /// Extract content file from a non-resource stream.
+        /// </summary>
+        /// <param name="stream">Stream to be extracted or decompiled.</param>
+        public static ContentFile ExtractNonResource(Stream stream)
+        {
+            var buffer = new byte[4];
```
xPaw, line 234:
```suggestion
            var buffer = stackalloc byte[4];
```

ValveResourceFormat/IO/FileExtract.cs
```diff
@@ -225,6 +225,34 @@ public static ContentFile Extract(Resource resource, IFileLoader fileLoader)
...
+            return magic switch
+            {
+                FlexSceneFile.FlexSceneFile.MAGIC => new FlexSceneExtract(stream).ToContentFile(),
+                _ => null,
+            };
+        }
+
+        public static bool IsNonResourceFile(string fileName)
+        {
+            return Path.GetExtension(fileName) == ".vfe";
```
xPaw, line 253: Realistically should just use magic bytes (merged with ExtractNonResource),

xPaw (approved):

muk-as:
VRF GUI can decompile vfe to txt
VRF Decompiler (with -d) return compiled vfe file
image

PR #766, opened 2024-04-11
Improperly implemented water

H7perus:
Trying to figure out why the texture just won't read.


Headcrabed:
Any updates?

H7perus:
Soon™. I have exams to go through. I wanted to do it properly in a bit from the ground up and closely follow the games water shader. Unless someone else wishes to accelerate this on their own.
This here was, so far, really hacky and also has issues with visible tiling. Better to at least use the games own water resources.

PR #779, opened 2024-04-25
Add mesh properties onto glTF extras when decompiling

IOExceptional:
Add mesh properties onto glTF extras when decompiling

Things like `targetname` and `classname` can now be seen from anything which uses the glTF models. 

This is helpful to determine dynamic entities and things that change between game modes (i.e. the extra barriers for wingman maps)

image


xPaw:
Seems simple enough. Does this get exposed in Blender or any other tools?

xPaw:
Is there something that prevents putting entity properties on the model itself (ModelRoot), and not meshes? Meshes are cached, so this could have issues (and multiple meshes will have same props).

PR #790, opened 2024-05-23
Post Processing (Tonemapping + Color Correction)

Purrspctiv:
Implements tonemapping and color correction!
Also has some corrections to the sky shader.

KNOWN ISSUES:
- Corrupted LUT on specific maps. Known maps include cs_office and most hlvr maps. No clue why this is happening, this is the main thing to investigate and solve.
- Viewmodes are WIP

The code isn't fully cleaned up, and it's not ready to be merged into master yet. There's still some things I need to take care of. The main purpose of this pull request (currently) is so others can look into the corrupted LUT issue, which is completely baffling me.


ValveResourceFormat/Resource/ResourceTypes/PostProcessing.cs
```diff
@@ -70,11 +70,11 @@ public byte[] GetRAWData()
             var j = 0;
             for (var i = 0; i < lut.Length; i++)
             {
-                // Skip each 4th byte
-                if (((i + 1) % 4) == 0)
-                {
-                    continue;
-                }
+                // Skip each 4th byte (this doesn't change anything)
```
xPaw, line 73: This makes the whole loop redundant. Does this still work correctly for the exporting?

GUI/Types/Renderer/ScenePostProcessVolume.cs
```diff
@@ -0,0 +1,257 @@
...
+                    }
+                    Log.Info(nameof(ScenePostProcessVolume), pixelValues);
+                }
+#endif
+
+                var buffer = ArrayPool<byte>.Shared.Rent(dimensions * dimensions * dimensions * bytesPerPixel);
+
+                try
+                {
+                    GL.TextureSubImage3D(ColorCorrectionLUT.Handle, 0, 0, 0, 0, dimensions, dimensions, dimensions, PixelFormat.Rgba, PixelType.UnsignedByte, buffer);
```
xPaw, line 228: A buffer is being rented, but nothing is written to it, so this writes garbage data to the texture (im surprised this worked at all on any map).

Did you mean to write `data` here without buffer?

Also keep in mind ArrayPool can return bigger sizes than requested.

GUI/Controls/GLViewerControl.cs
```diff
@@ -570,19 +582,34 @@ private void HandleResize()
...
+            }
+
+            if (FinalFramebuffer.InitialStatus == FramebufferErrorCode.FramebufferUndefined)
+            {
+                var status = FinalFramebuffer.Initialize();
+
+                if (status != FramebufferErrorCode.FramebufferComplete)
+                {
+                    Log.Error(nameof(GLViewerControl), $"PostProcess Framebuffer failed to bind with error: {status}");
+                    Log.Info(nameof(GLViewerControl), "Falling back to default framebuffer.");
```
xPaw, line 609: You can put this into the above message.

GUI/Types/Renderer/WorldLoader.cs
```diff
@@ -712,6 +714,102 @@ static bool IsCamera(string cls)
...
+                        Transform = transformationMatrix, // needed if model is used
+                    };
+
+                    var postProcessResourceFilename = entity.GetProperty<string>("postprocessing");
+
+                    if (postProcessResourceFilename != null)
+                    {
+                        var postProcessResource = guiContext.LoadFileCompiled(postProcessResourceFilename);
+
+                        if (postProcessResource != null)
```
xPaw, line 755: Should we bail out and not call `AddPostProcessVolume` if LUT fails to load?

xPaw:
Add a checkbox to disable color correction / post processing.

GUI/Types/Renderer/GLSceneViewer.cs
```diff
@@ -184,6 +185,16 @@ public virtual void PreSceneLoad()
 
             var defaultCubeTexture = GuiContext.MaterialLoader.LoadTexture(cubeFogResource);
             Textures.Add(new(ReservedTextureSlots.FogCubeTexture, "g_tFogCubeTexture", defaultCubeTexture));
+
+
+            // TODO: add backup for this
```
xPaw, line 190: You can see how that's done for the BRDF texture.

GUI/Types/Renderer/Shaders/post_processing.frag
```diff
@@ -0,0 +1,125 @@
...
+
+
+#if F_COLOR_CORRECTION_LUT
+
+layout(location=2) uniform sampler3D g_tColorCorrection;
+uniform float g_flColorCorrectionDefaultWeight;
+
+vec3 ApplyColorCorrection(vec3 vColor)
+{
+    vec3 ColorCorrectedColor = texture(g_tColorCorrection, saturate(vColor) * 0.9688 + 0.0156).rgb;
```
kristiker, line 84:
```suggestion
    uniform int g_nColorCorrectionSize = 32;
    const float x = (g_nColorCorrectionSize-1/g_nColorCorrectionSize);
    const float y = (1/(g_nColorCorrectionSize*2);
    vec3 ColorCorrectedColor = texture(g_tColorCorrection, saturate(vColor) * x +y).rgb;
```

GUI/Types/Renderer/Shaders/post_processing.frag
```diff
@@ -0,0 +1,125 @@
...
+
+    // Finally, convert from Linear to Gamma space
+    vec3 vTonemappedColorSRGB = SrgbLinearToGamma(vTonemappedColor);
+
+    return vTonemappedColorSRGB;
+}
+
+
+
+#if F_COLOR_CORRECTION_LUT
```
kristiker, line 77: I think we might be able to use this shader without any define, and use uniform bools if really necessary.

GUI/Controls/GLViewerControl.cs
```diff
@@ -372,7 +374,8 @@ private static void OnDebugMessage(DebugSource source, DebugType type, int id, D
         private static readonly DebugProc OpenGLDebugMessageDelegate = OnDebugMessage;
 
         public Framebuffer GLDefaultFramebuffer;
-        public Framebuffer MainFramebuffer;
+        public Framebuffer BasePassFramebuffer;
+        public Framebuffer FinalFramebuffer;
```
kristiker, line 378: Maybe we can postprocess directly from `MainFramebuffer` into the `GLDefaultFramebuffer`.

GUI/Types/Renderer/Shaders/post_processing.frag
Purrspctiv: Good point- funnily enough this is based on Valve's shader code, so I guess they just assumed it'll always be 32

GUI/Types/Renderer/Shaders/post_processing.frag
Purrspctiv: We'd have to provide a texture for g_tColorCorrection still though, right?

ValveResourceFormat/Resource/ResourceTypes/PostProcessing.cs
Purrspctiv: Fixed!

GUI/Types/Renderer/Shaders/post_processing.frag
xPaw: That might have been optimized by the compiler.

xPaw:
The vtex needs to be added as embedded resource here: https://github.com/ValveResourceFormat/ValveResourceFormat/blob/0486309021e6e3c875b42c94cc34d4598542754f/GUI/GUI.csproj#L41

GUI/Types/Renderer/Shaders/post_processing.frag
kristiker: Yeah weird that its hardcoded, but we can do better why not.

GUI/Types/Renderer/Shaders/post_processing.frag
kristiker: With a boolean check, maybe not.

(now I'm wondering what a passthrough LUT would be like lol)

GUI/Types/Renderer/Shaders/post_processing.frag
```diff
@@ -0,0 +1,96 @@
...
+    // So, we apply a 1-color-value dither to break up color banding.
+    // This is part of the original code and actually works extremely well.
+    vec3 subPrecisionDither = (blueNoise - 0.5) * g_vBlueNoiseDitherParams.w;
+    return vColor + subPrecisionDither;
+}
+
+vec4 SampleColorBuffer(vec2 coords)
+{
+    vec4 singleSampleColor = texelFetch(g_tColorBuffer, ivec2(coords.xy), int(gl_SampleID));
+    singleSampleColor.rgb = singleSampleColor.rgb / (max3(singleSampleColor.rgb) + 1.0);
```
kristiker, line 74: What's the divide by `max3(singleSampleColor.rgb) +1 ` doing

GUI/Types/Renderer/ScenePostProcessVolume.cs
kristiker: > im surprised this worked at all on any map

I'm guessing it rented same array as the one rented by binarykv3.

PR #796, opened 2024-05-27
vmap export _d suffix

Angel-foxxo:


GUI/Types/Exporter/ExportFile.cs
```diff
@@ -22,6 +22,11 @@ public static void ExtractFileFromPackageEntry(PackageEntry file, VrfGuiContext
 
         public static void ExtractFileFromStream(string fileName, Stream stream, VrfGuiContext vrfGuiContext, bool decompile)
         {
+            if (Path.GetExtension(fileName) == ".vmap_c")
```
xPaw, line 25: This should at least go inside of the `decompile` if.

ValveResourceFormat/IO/MapExtract.cs
```diff
@@ -1324,6 +1324,13 @@ private static Vector3 SrgbLinearToGamma(Vector3 vLinearColor)
 
         return vGammaColor;
     }
+    public static string AddSuffixToVmapName(string path)
+    {
+        var newVmapName = Path.GetFileNameWithoutExtension(path) + "_d" + Path.GetExtension(path);
+        var newVmapPath = NormalizePath(Path.Combine(Path.GetDirectoryName(path), newVmapName));
```
xPaw, line 1330: I think this line needs to be removed. In fact, I wouldn't put this method in map extract at all, but rather modify the variable going into `FileName = Path.GetFileNameWithoutExtension(fileName)` in the save dialog.

PR #806, opened 2024-06-06
add Texture.ReadTextureMipLevel()

in0finite:
I need to read only 1 mip level from texture, to reduce the memory required to load the map.

### DIFF
--- a/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
+++ b/ValveResourceFormat/Resource/ResourceTypes/Texture.cs
@@ -763,6 +763,22 @@ namespace ValveResourceFormat.ResourceTypes
             }
         }
 
+        public void ReadTextureMipLevel(Span<byte> output, uint mipLevel)
+        {
+            var bufferSize = CalculateBufferSizeForMipLevel(mipLevel);
+
+            if (output.Length != bufferSize)
+            {
+                throw new ArgumentException($"Buffer size ({output.Length}) should be equal to {bufferSize}, mip level {mipLevel}");
+            }
+
+            Reader.BaseStream.Position = Offset + Size;
+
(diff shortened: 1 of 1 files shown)


ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -763,6 +763,22 @@ public IEnumerable<(int Level, int Width, int Height, int BufferSize)> GetEveryM
             }
         }
 
+        public void ReadTextureMipLevel(Span<byte> output, uint mipLevel)
```
xPaw, line 766: Add xmldoc

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -763,6 +763,22 @@ public IEnumerable<(int Level, int Width, int Height, int BufferSize)> GetEveryM
...
+        {
+            var bufferSize = CalculateBufferSizeForMipLevel(mipLevel);
+
+            if (output.Length != bufferSize)
```
xPaw, line 770:
```suggestion
            if (bufferSize > output.Length)
```

Allow bigger array pools.

ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -618,7 +618,7 @@ public int CalculateTextureDataSize()
             return bytes;
         }
 
-        private int CalculateBufferSizeForMipLevel(uint mipLevel)
+        public int CalculateBufferSizeForMipLevel(uint mipLevel)
```
xPaw, line 621: Add xmldoc for this too then (also add the params to the other one)

PR #815, opened 2024-06-26
Fix vents_c `timesToFire` is not `1` & `-1` error

TianQiBuTian:
There are extremely few map authors who use special methods to ensure that the entity output is not fired only once or infinitely.
Video:

https://github.com/ValveResourceFormat/ValveResourceFormat/assets/26628474/012e21f7-651b-4f83-89a5-e244dbe54271

Before and after:
QQ截图20240627051011


kristiker:
Thanks, can you maybe improve the graphical output on map viewer as well? You can double click the entity to view the information.
 
Code is here:
https://github.com/ValveResourceFormat/ValveResourceFormat/blob/c9df08044b4448da23bb995b737ac2c78e66d72d/GUI/Forms/EntityInfoForm.cs#L107

PR #816, opened 2024-06-27
Improve the display of `timesToFire` in graphics

TianQiBuTian:
See #815 

After:
QQ截图20240628060313

### DIFF
--- a/GUI/Forms/EntityInfoForm.cs
+++ b/GUI/Forms/EntityInfoForm.cs
@@ -98,13 +98,28 @@ namespace GUI.Forms
             var delay = connectionData.GetFloatProperty("m_flDelay");
             var timesToFire = connectionData.GetInt32Property("m_nTimesToFire");
 
+            string stimesToFire = "";
+
+            switch (timesToFire)
+            {
+                case 1:
+                    stimesToFire = "Only Once";
+                    break;
+                case >= 2:
+                    stimesToFire = $"Only {timesToFire} Times";
+                    break;
+                default:
+                    stimesToFire = "Infinite";
+                    break;
+            }
+
             dataGridOutputs.Rows.Add(new string[] {
                 outputName,
                 targetName,
                 inputName,
                 parameter,
                 delay.ToString(NumberFormatInfo.InvariantInfo),
-                timesToFire == 1 ? "Yes" : "No"
+                stimesToFire
             });
         }
     }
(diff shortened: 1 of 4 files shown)


GUI/Forms/EntityInfoForm.cs
```diff
@@ -98,13 +98,31 @@ public void AddConnection(KVObject connectionData)
...
+                case 1:
+                    stimesToFire = "Only Once";
+                    break;
+                case 2:
+                    stimesToFire = "Only Twice";
+                    break;
+                case >= 3:
+                    stimesToFire = $"Only {timesToFire} Times";
+                    break;
+                default:
```
xPaw, line 114: Change this to case -1 so we know it's expected, and any other value would say `-2 times` (shouldn't ever happen anyway).

GUI/Forms/EntityInfoForm.cs
```diff
@@ -98,13 +98,31 @@ public void AddConnection(KVObject connectionData)
...
+                    stimesToFire = "Infinite Times";
```
xPaw, line 115:
```suggestion
                    stimesToFire = "Infinite";
```

GUI/Forms/EntityInfoForm.cs
```diff
@@ -98,13 +98,31 @@ public void AddConnection(KVObject connectionData)
...
             var timesToFire = connectionData.GetInt32Property("m_nTimesToFire");
 
+            string stimesToFire = "";
+
+            switch (timesToFire)
+            {
+                case 1:
+                    stimesToFire = "Only Once";
+                    break;
+                case 2:
```
xPaw, line 108: Why are we specializing 2 times into twice?

GUI/Forms/EntityInfoForm.cs
TianQiBuTian: Setting 0 is also infinite times. I am doing this to prevent the occurrence of `Only 0 Times` or `Only -2 Times`.

GUI/Forms/EntityInfoForm.cs
TianQiBuTian: Should I modify the following code again to make the two identical?
https://github.com/ValveResourceFormat/ValveResourceFormat/blob/25ac2246fed63e0321eaecef0e419e3fc0456aed/ValveResourceFormat/Resource/ResourceTypes/EntityLump.cs#L313-L314

GUI/Forms/EntityInfoForm.cs
kristiker: -2 also infinite?

xPaw: Case 0 is fine. I am not confident that -2 would be infinite in the engine.

TianQiBuTian: Watch the video in #815, which includes tests for `-2` and `0`.
And this "bug" OR "feature" already exists in the source 1 engine, both of which are infinite.

kristiker (approved):

PR #820, opened 2024-07-20
Allocate some buffers on stack

kristiker:
Meshoptimizer and vbib thing has not been tested.

### DIFF
--- a/ValveResourceFormat/Resource/Blocks/VBIB.cs
+++ b/ValveResourceFormat/Resource/Blocks/VBIB.cs
@@ -262,6 +262,9 @@ namespace ValveResourceFormat.Blocks
         {
             var result = new Vector2[vertexBuffer.ElementCount];
 
+            var offset = (int)attribute.Offset;
+            var data = vertexBuffer.Data.AsSpan();
+
             switch (attribute.Format)
             {
                 case DXGI_FORMAT.R32G32_FLOAT:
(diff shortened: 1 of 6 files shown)


ValveResourceFormat/Utils/StreamHelpers.cs
```diff
@@ -20,26 +20,20 @@ public static string ReadNullTermString(this BinaryReader stream, Encoding encod
...
 
-                int bytesRead;
-                var totalRead = 0;
-                while ((bytesRead = stream.Read(data, totalRead, characterSize - totalRead)) != 0)
-                {
-                    totalRead += bytesRead;
-                }
-
-                if (encoding.GetString(data, 0, characterSize) == "\0")
+                if (bytesRead == 0 || encoding.GetString(data) == "\0")
```
kristiker, line 31: Hm. If `bytesRead` is less than `characterSize`, data will have garbage value. For example it might not encode as NULL with UTF16.

xPaw:
Changes look fine to me if you tested all the codepaths you changed.

ValveResourceFormat/Resource/Blocks/VBIB.cs
```diff
@@ -436,26 +425,23 @@ public static ushort[] GetBlendIndicesArray(OnDiskBufferData vertexBuffer, Rende
...
                         for (var i = 0; i < vertexBuffer.ElementCount; i++)
                         {
-                            // TODO: Simplify this with Unsafe.ReadUnaligned
-                            Buffer.BlockCopy(vertexBuffer.Data, offset, shorts, 0, 4);
+                            var ushorts = MemoryMarshal.Cast<byte, ushort>(data.Slice(offset, count * sizeof(ushort)));
 
-                            System.Diagnostics.Debug.Assert(shorts[0] >= 0);
-                            System.Diagnostics.Debug.Assert(shorts[1] >= 0);
+                            System.Diagnostics.Debug.Assert(ushorts[0] >= short.MaxValue);
+                            System.Diagnostics.Debug.Assert(ushorts[1] >= short.MaxValue);
```
kristiker, line 442: feels like bad assert (should be <)

PR #833, opened 2024-08-26
extract some Explorer stuff to library

in0finite:


ValveResourceFormat/Utils/SteamLibraryGameInfo.cs
```diff
@@ -0,0 +1,4 @@
+namespace ValveResourceFormat.Utils
+{
+    public record struct SteamLibraryGameInfo(int AppID, string AppName, string SteamPath, string GamePath);
```
xPaw, line 3: Put this inside of GameFolderLocator

ValveResourceFormat/Utils/GameFolderLocator.cs
```diff
@@ -0,0 +1,158 @@
+using Microsoft.Win32;
+using System.Globalization;
+using System.IO;
+using System.Runtime.InteropServices;
+using ValveKeyValue;
+
+namespace ValveResourceFormat.Utils
```
xPaw, line 7: Let's put this into IO, not Utils.

ValveResourceFormat/Utils/GameFolderLocator.cs
```diff
@@ -0,0 +1,158 @@
...
+{
+    public static class GameFolderLocator
+    {
+        public static string GetSteamPath()
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
xPaw, line 13: We might as well support other OSes then: https://github.com/SteamDatabase/SteamTokenDumper/blob/c491265dd4d78f8bb0aac00320b8632c6f6a3f81/SteamClientData.cs#L200

PR #834, opened 2024-08-27
Theme System

LordOfXen:
A fully new theme system based on WinForm without changing any existing code (except new custom controls).
This does not touch any functionality of the GUI.

Multiple themes can be added (even by users) and can be changed at runtime through the context menu.
This is dark theme:
image
This is the white theme:
image
And users can choose to have custom themes:
image
**_Note that the image above shows an example as I just picked random colors to show the custom theme dialog._**

Some more screenshots:
image
image
image
image
image

For the dark theme I took the colors from Hammer 2 and for white theme I just picked on my own, feel free to change the default themes to any you wish.
If you'd like to have some changes on this before merging, feel free to check on my fork.


Angel-foxxo:
We already have a similar winforms themeing branch which also implements a custom title bar, but progress was halted until the release on .net 9 later this year which may or may not make our life easier

image

LordOfXen:
Yeah I have seen so, until that one is ready, this can be used too though.
Just having white right now is a pain at night.

xPaw:
Merging this would just produce more work (you just copypasted the entire gui as new files for one, and has the same issues we've been trying/waiting to fix), and if you've seen the existing work you should have communicated with us about that.

LordOfXen:
I disagree. This would not cause any more work as it changes basically nothing else than the names of classes on the existing code (e.g. Button to CustomButton). Other than these the rest are all new classes and can be shift-deleted when migrating to the new UI.

Angel-foxxo:
This also holds true for the existing PR, which we already put all this work into and then some to fix most of the pixel issue like misaligned icons and text, this is not needed.

LordOfXen:
Ah then that makes sense, I thought the existing PR wasn't even working. In this case, this is not needed.

LordOfXen:
How about if we move to Avalonia instead?

PR #846, opened 2024-10-13
Remove REDIBlock, and upconvert it to kv3 based Resource Edit Info

kristiker:


ValveResourceFormat/Resource/ResourceTypes/Texture.cs
```diff
@@ -797,14 +797,13 @@ public TextureCodec RetrieveCodecFromResourceEditInfo()
...
-            || !Resource.EditInfo.Structs.TryGetValue(ResourceEditInfo.REDIStruct.SpecialDependencies, out var specialDepsRedi))
+            if (Resource.EditInfo == null)
             {
                 return codec;
             }
 
-            var specialDeps = (SpecialDependencies)specialDepsRedi;
-            var textureCompilerDependencies = specialDeps.List.Where(dependancy => dependancy.CompilerIdentifier == "CompileTexture");
+            var textureCompilerDependencies = Resource.EditInfo.SpecialDependencies
+                .Where(dependency => dependency.CompilerIdentifier == "CompileTexture");
```
xPaw, line 806:
```suggestion
                .Where(static dependency => dependency.CompilerIdentifier == "CompileTexture");
```

ValveResourceFormat/Resource/ResourceTypes/Material.cs
```diff
@@ -146,10 +146,7 @@ private KVObject GetInputSignatureObject()
                 return null;
             }
 
-            var extraStringData = (ExtraStringData)Resource.EditInfo.Structs[ResourceEditInfo.REDIStruct.ExtraStringData];
-            var inputSignatureString = extraStringData.List.Where(x => x.Name == "VSInputSignature").FirstOrDefault()?.Value;
-
-            if (inputSignatureString == null)
+            if (Resource.EditInfo.SearchableUserData.FirstOrDefault(x => x.Key == "VSInputSignature").Value is not string inputSignatureString)
```
xPaw, line 149: Is this not a null ref?

ValveResourceFormat/Resource/Blocks/ResourceEditInfoStructs/SpecialDependency.cs
```diff
@@ -0,0 +1,31 @@
...
+using ValveResourceFormat.Serialization;
+using ValveResourceFormat.Serialization.KeyValues;
+
+namespace ValveResourceFormat.Blocks.ResourceEditInfoStructs
+{
+    public class SpecialDependency
+    {
+        public string String { get; set; }
+        public string CompilerIdentifier { get; set; }
+        public long Fingerprint { get; set; } // why long?
```
xPaw, line 12: Probably because GetIntegerProperty returns a long

ValveResourceFormat/Resource/ResourceTypes/Material.cs
kristiker: Return type is `KeyValuePair<string, object>` so a valuetype. `Value` accessor may be null though, and that is correctly checked.

ValveResourceFormat/Resource/Blocks/ResourceEditInfoStructs/SpecialDependency.cs
kristiker: I will make it a uint regardless, and add a throw in KVObject constructor if value exceeds uint max.

ValveResourceFormat/Resource/ResourceTypes/Material.cs
kristiker: as a value type, `default(KeyValuePair<string, object>>)` is not null

PR #875, opened 2024-12-10
Do not change the model BLENDINDICES stream

kristiker:
* Currently, we remap mesh bone indices to use indices from the model's skeleton. This adds:
  * Loading time and processing
  * Has overflow bugs. If the on disk VBIB is RGBA8888, and the remap table says index 0 corresponds to model bone 300, it overflows. (todo: fix this for exporting)
* A better approach is to adjust the indices on the animation texture so that each mesh can essentially have its own set of bones it can fetch starting from 0. The only drawback is that the animation texture becomes bigger, and many bones get duplicated.

Tested:
- [x] chicken.vmdl cs2
- [x] abrams.vmdl citadel
- [x] gman.vmdl hlvr
- [x] cs2, citadel, steamvr maps
- [x] ak47 glTF export

### DIFF
--- a/GUI/Types/Renderer/Shader.cs
+++ b/GUI/Types/Renderer/Shader.cs
@@ -130,16 +130,16 @@ namespace GUI.Types.Renderer
             }
         }
 
-        public void SetAnimationData(bool animated, int boneCount = 0, int weightCount = 0)
+        public void SetBoneAnimationData(bool animated, int boneOffset = 0, int boneCount = 0, int weightCount = 0)
         {
             var uniformLocation = GetUniformLocation("iAnimationData");
             if (uniformLocation > -1)
             {
-                GL.ProgramUniform4(Program, uniformLocation, animated ? 1u : 0u, (uint)boneCount, (uint)weightCount, 0);
+                GL.ProgramUniform4(Program, uniformLocation, animated ? 1u : 0u, (uint)boneOffset, (uint)boneCount, (uint)weightCount);
             }
         }
 
-        public void SetUniform4Array<T>(string name, int count, float[] value)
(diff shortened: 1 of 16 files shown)


GUI/Types/Renderer/Shader.cs
```diff
@@ -130,7 +130,16 @@ public void SetUniform4(string name, Vector4 value)
...
+        public void SetAnimationData(bool animated, int boneCount = 0, int weightCount = 0)
+        {
+            var uniformLocation = GetUniformLocation("iAnimationData");
+            if (uniformLocation > -1)
+            {
+                GL.ProgramUniform4(Program, uniformLocation, animated ? 1u : 0u, (uint)boneCount, (uint)weightCount, 0);
+            }
+        }
+
+        public void SetUniform4Array<T>(string name, int count, float[] value)
```
xPaw, line 142: wrong T?

GUI/Types/Renderer/ShapeSceneNode.cs
```diff
@@ -266,8 +266,6 @@ public override void Render(Scene.RenderContext context)
             GL.UseProgram(renderShader.Program);
 
             renderShader.SetUniform4x4("transform", Transform);
-            renderShader.SetUniform1("bAnimated", false);
-            renderShader.SetUniform1("sceneObjectId", Id);
```
xPaw, line 270: keep the ids

kristiker: I found these lines to be useless because the shader does not use bone animation at all. I guess it helps with `context.ReplacementShader`, weird design though.

PR #876, opened 2024-12-11
AnimationSystem2 support

kristiker:
Valve is developing a new animation system for Source 2 and has pushed a few resource files in Deadlock and CS2 that we can experiment with. Open source implementation that inspires this new Source 2 system can be found [here](https://github.com/BobbyAnguelov/Esoterica/blob/bed49a68d0dcf7d4863a8ab652e8825d7638a515/Code/Engine/Animation/AnimationClip.cpp#L22). 

#### TODO
- [ ] Parse animation sequence/clip (NmClip)
  - [x] Metadata
  - [x] Bone Positions
  - [x] Bone Rotations
  - [x] Bone Scales
- [x] Parse skeleton (NmSkel)
- [x] Test animations on renderer

### DIFF
--- a/GUI/Types/Renderer/GLModelViewer.cs
+++ b/GUI/Types/Renderer/GLModelViewer.cs
@@ -7,8 +7,6 @@ using GUI.Utils;
 using ValveResourceFormat.IO;
 using ValveResourceFormat.ResourceTypes;
 using ValveResourceFormat.ResourceTypes.ModelAnimation;
-using ValveResourceFormat.ResourceTypes.ModelAnimation2;
-
 
 #nullable disable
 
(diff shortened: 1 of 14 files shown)


GUI/Types/Renderer/GLModelViewer.cs
```diff
@@ -45,6 +50,11 @@ public GLModelViewer(VrfGuiContext guiContext, PhysAggregateData phys) : base(gu
             this.phys = phys;
         }
 
+        public GLModelViewer(VrfGuiContext guiContext, AnimationClip anim) : base(guiContext)
+        {
+            this.clip = anim;
+        }
+
```
kristiker, line 57: It might be better to make a different class. Maybe later on.

GUI/Types/Renderer/GLSceneViewer.cs
```diff
@@ -255,6 +255,11 @@ public virtual void PostSceneLoad()
                 offset = Math.Clamp(offset, 0f, 2000f);
                 var location = new Vector3(offset, 0, offset);
 
+                if (this is GLModelViewer { clip: not null })
```
xPaw, line 258: dont like this hack

PR #877, opened 2024-12-14
Fix KV3 text parser incorrectly adding double quotes when reading flagged value inside an array

KashimuraBaka:
Repro:
```
<!-- kv3 encoding:text:version{e21c7f3c-8a33-41c5-9977-a76d3a32aa0d} format:generic:version{7412167c-06e9-4698-aff2-e63eb59037e7} -->
{
 a =
 [
  resource:"path",
 ]
}
```


xPaw:
This whole function is some dank code. Can you add some test for this please?

PR #887, opened 2025-01-13
Always keep a `Resource` reference in all blocks, remove resource argument on `Block.Read`

kristiker:


ValveResourceFormat/Resource/Resource.cs
```diff
@@ -232,59 +232,55 @@ public void Read(Stream input, bool verifyFileSize = true)
...
 
-                switch (block.Type)
+                if (block.Type is BlockType.RED2 or BlockType.REDI or BlockType.NTRO)
                 {
-                    case BlockType.REDI:
-                    case BlockType.RED2:
-                        block.Read(Reader, this);
+                    block.Read(Reader);
 
+                    if (block.Type is BlockType.RED2 or BlockType.REDI)
```
xPaw, line 243: feels like pointless nest, just make it two separate ones

kristiker: Done

PR #908, opened 2025-03-28
add bone_parent property to exported ragdoll collider nodes

### LINKED ISSUE #907: Parent bone link for ragdoll parts is lost when decompiling vmdl_c file

Decompiling a `.vmdl_c` file creates physics shape nodes with correct shape parameters, but no `parent_bone` property, requiring manual setup when working with the decompiled model.

The decompiler should set the parent bone automatically.

The relevant bone names appear to be stored in the `m_boneNames` array, mapped to each other by index, i.e. the shapes in the `i`-th `m_parts` element belong to the bone with the name at the `i`-th `m_boneNames` element.

_No response_

zirrboy:
closes #907

### DIFF
--- a/ValveResourceFormat/IO/ModelExtract.ValveModel.cs
+++ b/ValveResourceFormat/IO/ModelExtract.ValveModel.cs
@@ -684,12 +684,19 @@ partial class ModelExtract
 
         if (physAggregateData is not null)
         {
-            foreach (var physicsPart in physAggregateData.Parts)
+            var boneNames = physAggregateData.Data.GetArray<string>("m_boneNames");
+            boneNames ??= [];
+
+            for (var i = 0; i < physAggregateData.Parts.Length; i++)
             {
+                var physicsPart = physAggregateData.Parts[i];
+                var parentBone = boneNames.Length > i ? boneNames[i] : string.Empty;
+
                 foreach (var sphere in physicsPart.Shape.Spheres)
                 {
                     var physicsShapeSphere = MakeNode(
                         "PhysicsShapeSphere",
+                        ("parent_bone", parentBone),
                         ("surface_prop", PhysicsSurfaceNames[sphere.SurfacePropertyIndex]),
                         ("collision_tags", string.Join(" ", PhysicsCollisionTags[sphere.CollisionAttributeIndex])),
                         ("radius", sphere.Shape.Radius),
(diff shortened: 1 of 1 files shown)


ValveResourceFormat/IO/ModelExtract.ValveModel.cs
```diff
@@ -684,12 +684,17 @@ Lazy<KVObject> MakeLazyList(string className)
 
         if (physAggregateData is not null)
         {
-            foreach (var physicsPart in physAggregateData.Parts)
+            var boneNames = this.physAggregateData.Data.GetArray<string>("m_boneNames");
+
+            for (global::System.Int32 i = 0; i < physAggregateData.Parts.Length; i++)
```
kristiker, line 689: var i = 0

ValveResourceFormat/IO/ModelExtract.ValveModel.cs
```diff
@@ -684,12 +684,17 @@ Lazy<KVObject> MakeLazyList(string className)
 
         if (physAggregateData is not null)
         {
-            foreach (var physicsPart in physAggregateData.Parts)
+            var boneNames = this.physAggregateData.Data.GetArray<string>("m_boneNames");
```
kristiker, line 687: The list could be empty. You might get away with adding an empty string as the `parent_bone` property if that's the case.

ValveResourceFormat/IO/ModelExtract.ValveModel.cs
```diff
@@ -684,12 +684,18 @@ Lazy<KVObject> MakeLazyList(string className)
...
+
+            for (var i = 0; i < physAggregateData.Parts.Length; i++)
             {
+                var physicsPart = physAggregateData.Parts[i];
+
                 foreach (var sphere in physicsPart.Shape.Spheres)
                 {
                     var physicsShapeSphere = MakeNode(
                         "PhysicsShapeSphere",
+                        ("parent_bone", boneNames.ElementAtOrDefault(i) ?? ""),
```
xPaw, line 698: Do we expect bones array to be less than parts?

I'd probably do `boneNames.Length > 0 ? boneNames[i] : string.Empty` (or null check if it cant be an empty array in the file.

zirrboy: I don't know of any cases other than 0 or matching length. I did it this way so it doesn't crash even if that assumption turns out wrong, but I can chance it.

zirrboy: Mixing parented and parentless shapes leads to an empty string entry in the bone names when using CSDK 10.

kristiker: Empty is expected, otherwise should have matching length. I also have a feeling that the whole property could be missing on older files, so the null check is ok.

PR #914, opened 2025-04-04
extract node creation utils from ModelExtract into a separate class

zirrboy:
A minor refactor to take a bit of load off the huge module and make the utility functions more readily accessible to other extraction code.


ValveResourceFormat/IO/NodeCreationHelpers.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return MakeValue(list);
+    }
+
+    internal static void AddItem(KVObject node, KVObject item)
+    {
+        Debug.Assert(node.IsArray);
+        node.AddProperty(null, MakeValue(item));
+    }
```
kristiker, line 58: This could be moved to KVObject itself

kristiker (approved):

PR #924, opened 2025-04-29
Fix model decompile BodyGroupChoice cannot correctly classify Mesh

KashimuraBaka:


xPaw:
Can you show a model that has this issue?

PR #929, opened 2025-05-03
Enable nullability in ValveMap.cs

Angel-foxxo:


ValveResourceFormat/IO/ContentFormats/ValveMap.cs
```diff
@@ -364,9 +362,9 @@ internal class CDmePolygonMeshDataStream<T> : DMElement
     public int SemanticIndex { get; set; }
     public int VertexBufferLocation { get; set; }
     public int DataStateFlags { get; set; }
-    public DMElement SubdivisionBinding { get; set; }
+    public DMElement? SubdivisionBinding { get; set; }
```
kristiker, line 365: If the vmap is not valid with these field as null, then this nullable annotation isn't really helping us here..

Angel-foxxo: I don't think it matters so much, it's more just telling it that this CAN be null and it's fine so it stops complaining, can't think of a batter way to do it

PR #951, opened 2025-07-05
Shader validator

kristiker:


Misc/ShaderValidator/ShaderValidator.cs
```diff
@@ -0,0 +1,13 @@
+
+internal class Program
+{
+    public static int Main(string[] args)
+    {
+        var progressReporter = new Progress<string>(Console.WriteLine);
```
xPaw, line 6: I think you can use Log.Info here

.github/workflows/validate_shaders.yml
```diff
@@ -0,0 +1,43 @@
+name: Validate
+
+on:
+  push:
+    paths:
+      - GUI/Types/Renderer/Shaders/**
+  pull_request:
+    paths-ignore:
```
xPaw, line 8:
```suggestion
    paths:
```

xPaw:
19bfe81ca7304c67c3c1d5dc002a79ec6ff430ed

PR #953, opened 2025-07-10
debug info improvements + text renderer improvements

Angel-foxxo:


GUI/Types/Renderer/SelectedNodeRenderer.cs
```diff
@@ -34,6 +37,107 @@ public SelectedNodeRenderer(Scene scene, TextRenderer textRenderer) : base(scene
...
+                {
+                    minDistance = distance;
+                    closestIndex = i;
+                }
+            }
+
+            return vertexInfos[closestIndex];
+        }
+
+        private Dictionary<SceneNode, VertexInfo[]> NodeVertexInfos = [];
```
xPaw, line 138: I don't see why this needs to be cached.

GUI/Types/Renderer/SelectedNodeRenderer.cs
```diff
@@ -229,6 +356,33 @@ public override void Render(Scene.RenderContext context)
...
+                    NodeClosestVertexInfo.Add(node, closestVertexInfo);
+                    UpdateBuffer();
+                }
+
+                var newClosestInfo = GetClosestVertexInfo(context.Camera, vertexInfos);
+
+                if (closestVertexInfo != newClosestInfo)
+                {
+                    NodeClosestVertexInfo[node] = newClosestInfo;
+                    UpdateBuffer();
```
xPaw, line 383: This class already contains UpdateEveryFrame bool for particles, so with the dynamic drawing here, we probably always want to recalculate the buffer then.

GUI/Types/Renderer/GLModelViewer.cs
```diff
@@ -331,6 +344,68 @@ protected override void OnPaint(object sender, RenderEventArgs e)
...
+            if (ModelStatsCheckbox.Checked)
+            {
+                selectedNodeRenderer.SelectNode(modelSceneNode);
+
+                using (new GLDebugGroup("Model Viewer Stats Text"))
+                {
+                    var meshesString = "";
+
+                    var modelMeshCount = 0;
+                    foreach (var embeddedMesh in model.GetEmbeddedMeshesAndLoD())
```
xPaw, line 357: This is horrible to call every draw, also doesn't work for separate mesh files. You should access the existing scene nodes.

PR #954, opened 2025-07-13
Outline rendering for selected objects

kristiker:
image

### DIFF
--- a/GUI/Types/Renderer/SelectedNodeRenderer.cs
+++ b/GUI/Types/Renderer/SelectedNodeRenderer.cs
@@ -180,7 +180,7 @@ namespace GUI.Types.Renderer
             {
                 var nodeName = node.Name ?? node.GetType().Name;
 
-                if (node is not SimpleBoxSceneNode)
+                if (node is not SimpleBoxSceneNode and not SpriteSceneNode)
                 {
                     AddBox(vertices, node.Transform, node.LocalBoundingBox, Color32.White, showSize: true);
                 }
--- a/GUI/Types/Renderer/SelectedNodeRenderer.cs
+++ b/GUI/Types/Renderer/SelectedNodeRenderer.cs
@@ -44,15 +44,18 @@ namespace GUI.Types.Renderer
             if (selectedNode >= 0)
             {
                 selectedNodes.RemoveAt(selectedNode);
+                node.IsSelected = false;
             }
             else
             {
                 selectedNodes.Add(node);
+                node.IsSelected = true;
             }
         }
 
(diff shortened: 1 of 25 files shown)


GUI/Types/Renderer/SelectedNodeRenderer.cs
```diff
@@ -150,7 +154,7 @@ int ClosestVertexInView(ReadOnlySpan<Vector3> vertices)
                         Color = axisColor,
                         Text = length.ToString("0.##", CultureInfo.InvariantCulture),
                         Center = true
-                    });
+                    }, fixedScale: false);
```
kristiker, line 157: Revert or adjust so it doesnt fill the screen so much

GUI/Types/Renderer/SelectedNodeRenderer.cs
```diff
@@ -63,6 +66,7 @@ public void SelectNode(SceneNode? node, bool forceDisableDepth = false)
             }
 
             selectedNodes.Add(node);
+            node.IsSelected = true;
```
kristiker, line 69: This class could fill the renderlists directly? It would need to check if the node passed frustum cull.

GUI/Types/Renderer/ShaderParser.cs
```diff
@@ -252,17 +252,31 @@ void LoadShaderString(string shaderFileToLoad, string? parentFile, bool isInclud
         }
 
 #if !DEBUG
+        public static bool ShaderFileExists(string name)
+        {
+            var resourceName = $"{ShaderDirectory}{name.Replace('/', '.')}";
+            var assembly = System.Reflection.Assembly.GetExecutingAssembly();
+            return assembly.GetManifestResourceInfo(resourceName) != null;
+        }
+
```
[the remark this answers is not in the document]
xPaw: I change it to get, it still throws?

GUI/Types/Renderer/Shaders/outline.geom
```diff
@@ -0,0 +1,40 @@
+#version 460
+
+layout(triangles) in;
+layout(triangle_strip, max_vertices = 21) out;
+
+#include "common/ViewConstants.glsl"
+
+uniform float g_flLineSize = 0.3;
```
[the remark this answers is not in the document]
xPaw: I think we just need to tweak the rendering a bit to make it more consistent, then no setting needed for this.

GUI/Types/Renderer/SelectedNodeRenderer.cs
xPaw: I mean we already fill `selectedNodes`, could use that?

PR #987, opened 2025-08-29
Add NmSkel decompilation and extend NmClip decompilation

LionDoge:
I tested the NmClip export on a handful of assets, but I didn't try every implemented event type yet.

Side note: DMX export seems to be not working fully, at least for skeletal animations. Directly exported file neither works in-game or for external tools (like Blender Source Tools) so users need to find some other way to export the animation sequences themselves for now. I thought about creating an issue on Datamodel.NET repo, but there seems to be a similar issue already, and it wasn't updated for 9 years at least. It seems that changing version to binary 9 for the DMX export will work in Blender Source Tools, but will crash the game, I haven't changed it here though.

### DIFF
--- a/ValveResourceFormat/IO/NmClipExtract.cs
+++ b/ValveResourceFormat/IO/NmClipExtract.cs
@@ -60,10 +60,11 @@ public class NmClipExtract
             var docEventTrack = BuildDocEventBasedOnEventClass(ev, ev.GetStringProperty("_class"));
             var startTimeSeconds = ev.GetFloatProperty("m_flStartTimeSeconds");
             var durationSeconds = ev.GetFloatProperty("m_flDurationSeconds");
+            var eventList = docEventTrack.GetArray<KVObject>("m_events").First();
             // Doc file event time stamps are given in frames they can be technically floats, but based on recompilation tests
             // these seem inconsistent, unless they're floored to int, then it matches up.
-            docEventTrack.GetArray<KVObject>("m_events").First().AddProperty("m_flStartTime", Math.Floor(startTimeSeconds * animation.Fps));
-            docEventTrack.GetArray<KVObject>("m_events").First().AddProperty("m_flDuration", Math.Floor(durationSeconds * animation.Fps));
+            eventList.AddProperty("m_flStartTime", Math.Floor(startTimeSeconds * animation.Fps));
+            eventList.AddProperty("m_flDuration", Math.Floor(durationSeconds * animation.Fps));
             docEventTracks.AddItem(docEventTrack);
         }
         kv.AddProperty("m_eventTracks", docEventTracks);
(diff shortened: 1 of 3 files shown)


ValveResourceFormat/IO/NmClipExtract.cs
```diff
@@ -0,0 +1,221 @@
...
+    // Returns a full event track.
+    private static KVObject BuildDocEventBasedOnEventClass(KVObject kvCompiledEvent, string className)
+    {
+        // From testing one event track in doc seems to correspond to one event in compiled asset
+        // even though m_events is an array inside each track.
+        var kvDocEventTrack = new KVObject(null);
+        var kvDocEvent = new KVObject("m_event");
+        kvDocEventTrack.AddProperty("m_type", "Duration");
+        kvDocEventTrack.AddProperty("m_bIsSyncTrack", kvCompiledEvent.ContainsKey("m_syncID"));
+        switch (className)
```
xPaw, line 83: I notice that most properties are directly copied from the source, what if all properties were copied, and only ones that need to be renamed were renamed?

Does this actually map all events, or are there more? Would be nice if we didn't have to manually maintain the full mapping.

kristiker: Yeah we do similar thing in other decompilers such as AnimationGraph.

You can extract class name by removing "CNm" and "Event", and create "CNmClipDocEvent_{class}". Pass through all properties by default, and only handle any special ones.

LionDoge:
Sorry for the delay, I was focusing on other things. However I applied your suggestion about making it more generic. I don't know if it's the best way to do it though. I split the only special cases in a switch statement so it's easier to add more if needed.

The schemas for events can be found in animdoclib and animlib modules from schema. I think dumpsource2 doesn't dump the resourcecompiler info, but this one does if you want to see yourself: https://github.com/GAMMACASE/Source2SchemaDumper

I haven't fully tested this yet by the way. Only on like two files so far, I will try to very soon and fix any potential issues.
You can also add this to assettypes_common.txt to test the file recompilation:
```
animgraph2_skeleton = 
        {
            _class = "CResourceAssetTypeInfo"
            m_FriendlyName = "Animation 2 Skeleton"
            m_Ext = "vnmskel"
            m_CompilerIdentifier = "CompileNmSkeleton"
            m_IconLg = "game:tools/images/animation/vnmskel_lg.png"
            m_IconSm = "game:tools/images/animation/vnmskel_sm.png"
            m_Blocks = 
            [
                {
                    m_BlockID = "DATA"
                    m_Encoding = "RESOURCE_ENCODING_KV3"
                },
            ]
        }
        animgraph2_clip = 
        {
            _class = "CResourceAssetTypeInfo"
            m_FriendlyName = "Animation 2 Clip"
            m_Ext = "vnmclip"
            m_CompilerIdentifier = "CompileNmClip"
            m_CompileDependsOnResourceTypes = [ "vnmskel" ]
            m_IconLg = "game:tools/images/animation/vnmclip_lg.png"
            m_IconSm = "game:tools/images/animation/vnmclip_sm.png"
            m_Blocks = 
            [
                {
                    m_BlockID = "DATA"
                    m_Encoding = "RESOURCE_ENCODING_KV3"
                },
            ]
        }
```

kristiker (approved):
Looks good, applied a few modifications myself.

Issue #789, opened 2024-05-19
Add plane and cube shapes to material viewer


zndxcvbn:
### Is your feature request related to a problem? Please describe.

This innovation will help you better see .vmat textures and other texture representations

### Describe the solution you'd like

Add plane and cube shapes, and also add a button to the interface where you can select a shape to view the .vmat texture

xPaw:
I wonder how we could detect which shape would be best to show by default for a given material.

Also doesn't a cube practically cover the same use case as a cube? Perhaps we could just show both a cube and a sphere?

xPaw:
To add to this, some models have a "PreviewModel"

For example gloves/paints/bloodhound_snakeskin_brass.vmat_c has this in DATA:
```
    m_stringAttributes = 
    [
        {
            m_name = "PreviewModel"
            m_value = "characters/models/shared/arms/glove_brokenfang/v_glove_brokenfang.vmdl"
        },
    ]
```

PR #1015, opened 2025-10-31
Material viewer box preview model

THG327:
Changes the preview model for the material viewer to use a box model (sourced from CS2 "core" VPK)
This has the advantage of displaying the whole texture clearly and being nicer to look at.


xPaw:
Reference: #789

It would be nice to be able to switch between them with a button

Angel-foxxo:
> Reference: #789
> 
> It would be nice to be able to switch between them with a button

it would be nice to have all the ones from source2, and the plane one should also position your camera above the plane

kristiker:
Box and Quad shapes are so simple that it is better to make some constructs in RenderableMesh to create them with just code. I'll give his a try.

PR #1025, opened 2025-11-09
pull material loader out of context and make it static

Angel-foxxo:


xPaw:
The cached textures hold GL objects, thus can't be cached across contexts.

PR #1044, opened 2025-12-16
Only decompile the highest available shader model

dr3fty:
SM can probably be removed from the decompiled file name


kristiker:
Messy but works

PR #1060, opened 2026-01-19
Attempt to improve AG1 decompilation

dirtkiller23:
Added random node positions ( Positions should use logic similar to AG2 graph viewer)
Added components (except Demo Settings)
Improve m_paramindex / m_tagindex
Add sequence tag/param spans
Add clipdatamanager
Improve state machine transitions, add random state positions, add state machine condition scripts
Added a lot of nodes


dirtkiller23:
Added add/subtract, bonemask and blend 1D nodes support

dirtkiller23:
All nodes are done except Motion Matching and State Machine (Condition scripts)

dirtkiller23:
Added State Machine Node / State Machine Component condition scripts.

PR #1064, opened 2026-01-23
Add export context menu for shader bytecode

dr3fty:


kristiker:
Thanks, its working.

image

PR #1073, opened 2026-01-29
Depth of field effect

kristiker:
image

image

### DIFF
--- a/GUI/Types/GLViewers/GLWorldViewer.cs
+++ b/GUI/Types/GLViewers/GLWorldViewer.cs
@@ -1,7 +1,6 @@
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-using System.Security.Cryptography;
 using System.Windows.Forms;
 using GUI.Controls;
 using GUI.Forms;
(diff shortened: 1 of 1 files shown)


Renderer/Renderer/DOFRenderer.cs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public void SetShaderParams()
+    {
+        var newParams = CreateInputParams();
+
+        if (CurrentDofParams.Equals(newParams))
+        {
+            return;
+        }
```
kristiker: Recalculation seems dependent on framebuffer width too. Also not the whole DofParams struct, but just MaxBlurSize, and RadScale.

Renderer/Renderer/DOFRenderer.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        CurrentDofParams = newParams;
+
+        var g_vInvRenderTargetDim = new Vector2(1.0f / DOFFrameBuffer!.Width, 1.0f / DOFFrameBuffer.Height);
+
+        DOF!.SetUniform1("flMaxBlurSize", CurrentDofParams.MaxBlurSize);
+
+        float[] offsets = new float[MAX_DOF_SAMPLES * 4];
```
kristiker: 4096 bytes, this can be stack allocated

PR #1098, opened 2026-02-24
Animgraph 2 System

kristiker:
Run animation graphs.

Based on animgraph system from [Esoterica Engine](https://github.com/BobbyAnguelov/Esoterica/), which is pretty much what the new AnimGraph 2 is based on.

Node classes were generated using the output from schema system dumper. See https://s2v.app/SchemaExplorer/cs2/animlib for more context.

todo:
- [ ] Use json instead of hpp files as base.

https://github.com/user-attachments/assets/260523e1-3570-43e7-b58e-4a0865f538a7

Very WIP.


kristiker:
Added clip based playback support #1106

PR #1104, opened 2026-03-06
Initial barn and omni2 light support

dr3fty:
Known issues / missing features:
- No tile culling
- Static 8k barn light shadow atlas
- Performance (both shadow rendering and light rendering itself is "not great")
- No alpha tested depth pass

### DIFF
--- a/Renderer/Renderer.cs
+++ b/Renderer/Renderer.cs
@@ -486,8 +486,9 @@ public class Renderer
         using var _ = new GLDebugGroup("Barn Light Shadows");
         Debug.Assert(BarnLightShadowBuffer != null);
 
-        // Reverse Z without flipping the projection matrix
-        GL.DepthRange(1.0, 0.0);
+        GL.DepthFunc(DepthFunction.Lequal);
+        GL.DepthRange(0.0, 1.0);
+        GL.ClearDepth(1.0);
         GL.FrontFace(FrontFaceDirection.Cw);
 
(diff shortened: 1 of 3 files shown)


Renderer/Renderer/WorldLightingInfo.cs
```diff
@@ -240,42 +277,267 @@ void AddLight(SceneLight light, uint index)
 
         public void StoreLightMappedLights_V2(List<SceneLight> lights)
```
xPaw: just remove the wrapper method

Renderer/Renderer/Scene.cs
```diff
@@ -737,28 +744,82 @@ public void SetupSceneShadows(Camera camera, int shadowMapSize)
...
+            var barnLightFrustumHash = lightFrustum.GetHashCode();
+            ref var entry = ref CollectionsMarshal.GetValueRefOrAddDefault(light.FaceShadowCache, faceIndex, out _);
+
+            if (entry.FrustumHash == barnLightFrustumHash && entry.DrawCalls is not null)
+            {
+                return;
+            }
+
+            entry.DrawCalls ??= CreateDepthOnlyDrawCallCollection();
+            CollectShadowDrawCalls(lightFrustum, includeStatic: true, includeDynamic: true, entry.DrawCalls);
```
kristiker: Cached results should be cleared when octree gets dirty. Actually caching makes no sense for dynamic objects... But it makes cpu time increase a lot from inefficient octree queries.

PR #1105, opened 2026-03-11
Fix UV export for decal/overlay meshes

ifpCD:
CreateVertexBufferAccessors numbers TEXCOORD/COLOR accessors locally per vertex buffer, so each buffer starts from TEXCOORD_0.

When a draw call uses multiple vertex buffers (geometry + a separate lightmap UV buffer) the second buffer's TEXCOORD_0 would silently overwrite the first's on the primitive which would replace the primary UV with the lightmap UV.

Before:
blender_m2te5s4RJg

After:
blender_M29IUyEUrJ


xPaw:
Hmm, this should already be handled in `CreateVertexBufferAccessors`, is that broken?

EDIT: To answer my question: it's because of separate vertex buffers.

PR #1124, opened 2026-03-28
Fix envmap render mode in vr_standard

Arenovas:
After the envmap rendering was fixed for the standard rendering mode with the vr_standard shader, it was still not fixed for the cubemap rendering mode. This removes the forced color space conversion when using said rendering mode so SteamVR Home cubemaps aren't blown out.
I only changed it for the vr_standard shader but should I also change the other shaders to mimic this behavior as well?

### DIFF
--- a/Renderer/Shaders/vr_standard.frag.slang
+++ b/Renderer/Shaders/vr_standard.frag.slang
@@ -274,8 +274,7 @@ void main()
     }
     else if (g_iRenderMode == renderMode_Cubemaps)
     {
-        vec3 viewmodeEnvMap = GetEnvironment(mat).rgb;
-        outputColor.rgb = SrgbLinearToGamma(viewmodeEnvMap);
+        outputColor.rgb = GetEnvironment(mat).rgb;
     }
     else if (g_iRenderMode == renderMode_Tint)
     {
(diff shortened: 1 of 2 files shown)


kristiker:
Debug render modes output srgb space colors, and the cubemap is sampled in linear space so this was correct no? In CS2 the cubemaps are definitely not blown out.

Arenovas:
Yeah in all the other S2 games the cubemap render mode looked proper but in steamvr home maps they looked blown out.
This change makes fixes the issue for those maps while leaving the rest untouched. HL:A maps are also unaffected by the change I made due to the earlier cubemap color space fix.

Unless the way it was output in the debug render mode was on purpose? Just seems odd to have them blown out in steamvr home maps.

kristiker:
I believe IBL and irradiance uses RGBM6, so you can rebase branch over 2f476b5314304fd54f6ffe62df5105852a45338a and decode inside GetEnvironment etc.

Arenovas:
Alright, I'll have a little poke around when I have a bit more time and see what I can do!

Arenovas:
So I managed to get the cubemaps and lightprobe textures rendering properly in their previews, although I have no clue if it should have been done in a different way.
It doesn't seem to have broken anything though on a quick look anyways.

PR #1148, opened 2026-04-27
Add and fix some shaders for SteamVR and Robot Repair

Arenovas:
I went and used the vr_standard shader to create a vr_simple shader to fix up the bugged materials in Robot Repair, and might also fix some materials in SteamVR as well in any cases that use the vr_simple shader.
I also added the vr_unlit shader in SteamVR to use the vr_black_unlit shader to fix some bugged materials in the City 17 Alleyway map.
Although now that I think about it, there might be materials that use the vr_unlit shader that aren't black, gotta go verify that but either way it'll either remain the same or maybe a new unlit shader would need to be made for it?


Renderer/Shaders/vr_unlit.vert.slang
```diff
@@ -0,0 +1,51 @@
...
+#include "common/compression.slang"
+
+centroid out vec4 vVertexColorOut;
+out vec3 vFragPosition;
+out vec3 vNormalOut;
+out vec3 vTangentOut;
+out vec3 vBitangentOut;
+out vec2 vTexCoordOut;
+
+uniform vec3 g_vColorTint = vec3(1.0);
```
kristiker:
```suggestion
uniform vec3 g_vColorTint = vec3(1.0); // SrgbRead(true)
```

Since multiplication below is being done in linear space. (and tints are usually linear)

Renderer/Renderer/Shaders/ShaderLoader.cs
```diff
@@ -417,7 +417,9 @@ public static string ShaderNameFromPath(string shaderFilePath)
             "tools_sprite.vfx" => "sprite",
             "global_lit_simple.vfx" => "global_lit_simple",
             "vr_black_unlit.vfx" or "csgo_black_unlit.vfx" => "vr_black_unlit",
+            "vr_unlit.vfx" => "vr_unlit",
             "vr_standard.vfx" => "vr_standard",
+            //"vr_simple.vfx" => "vr_simple", //Disabled until per appid shader support is added
```
kristiker: We have to add that on same PR. Its better to move this to another branch.

kristiker:
If there is any GameVfx vr_unlit traces in complex, please remove those too.

PR #1149, opened 2026-04-28
*.vnmgraph files export support

STL1te:
Implementation of vnmgraph file export, written mostly through vibe coding.


ValveResourceFormat/IO/NmGraphExtract.cs
```diff
@@ -0,0 +1,3454 @@
...
+    {
+        var vector = KVObject.Array();
+        vector.Add(source is not null && source.Count > 0 ? source[0] : 0.0f);
+        vector.Add(source is not null && source.Count > 1 ? source[1] : 0.0f);
+        vector.Add(source is not null && source.Count > 2 ? source[2] : 0.0f);
+        return vector;
+    }
+
+    private static float RadiansToDegrees(float radians)
+        => radians * (180.0f / MathF.PI);
```
kristiker: there is built in function in `float` class for this

xPaw:
This pr has the same problem as ag1 exporter of having thousands of lines of code with manually mapping each key. Anything can be done about that?

STL1te:
> This pr has the same problem as ag1 exporter of having thousands of lines of code with manually mapping each key. Anything can be done about that?

The code is now split into smaller partials for the main graph builder, variations, helper types, node specs, and virtual parameter type inference.
For the simpler nodes the manual switch cases were moved into declarative specs: class name, pins, input connections, and small extra fields live in one compact table now.

There is still some manual mapping left for complex nodes like state machines, transitions, blend nodes, IK, variation data, etc. 
I don’t think those can be cleanly generated from schemas alone because the schemas describe the runtime/document fields, but not the compiled-to-editable graph reconstruction rules.

But the repetitive part is now much easier to maintain and the remaining manual code is more isolated.

kristiker:
We probably dont need to upload these test resources in CI.

Lets do something like AG1 in [Tests/GameExtractTests.cs](https://github.com/ValveResourceFormat/ValveResourceFormat/blob/master/Tests/GameExtractTests.cs). Read from local game files and compare to a decompiled vnmgraph that you can upload.

The test should be minimal. The countless hardcoded `string.Contains` are unmaintainable.

STL1te:
> We probably dont need to upload these test resources in CI.
> 
> Lets do something like AG1 in [Tests/GameExtractTests.cs](https://github.com/ValveResourceFormat/ValveResourceFormat/blob/master/Tests/GameExtractTests.cs). Read from local game files and compare to a decompiled vnmgraph that you can upload.
> 
> The test should be minimal. The countless hardcoded `string.Contains` are unmaintainable.

I changed the tests to split the coverage into two parts:
- The live game test now follows the AG1-style approach: it reads vnmgraph files from a local CS2 install when available and checks that extraction does not fail.
This is meant to catch newly added/unsupported node types in real game files.
- The uploaded compiled test resources are kept as a small regression fixture batch based on viewmodel_gun.vnmgraph_c and its variation files.
Those files are needed because variation resolution depends on sibling compiled resources and this lets us test that behavior deterministically in CI without requiring a local game install.

ValveResourceFormat/IO/NmGraphExtract.cs
```diff
@@ -0,0 +1,2862 @@
...
+
+    private static KVObject MakeVector2(float x, float y)
+    {
+        var vector = KVObject.Array();
+        vector.Add(x);
+        vector.Add(y);
+        return vector;
+    }
+
+    private static string MakeGuid(string seed)
```
dirtkiller23:
```suggestion
    private static string MakeGuid(string seed) => Guid.NewGuid().ToString();
```
Could this method
```
private static string MakeGuid(string seed)
{
    var bytes = Encoding.UTF8.GetBytes(seed);
    var hash = SHA256.HashData(bytes);
    return new Guid(hash[..16]).ToString();
}
```
be changed to a 

```
private static string MakeGuid(string seed) => Guid.NewGuid().ToString();
```

Previous method ends up generating same GUIDs across exported multiple files which causes issues in some cases.

dirtkiller23 (changes requested):

ValveResourceFormat/IO/NmGraphExtract.cs
STL1te: The deterministic GUIDs are intentional here because some IDs are referenced before the corresponding node is emitted. For example, parameter reference nodes use MakeGuid($"root-control:{nodeIndex}") as a fallback for m_parameterUUID, and the actual root parameter node later gets the same ID from the same seed. If MakeGuid always returned a random GUID, those references could point to non-existent parameter nodes depending on traversal order.

STL1te: For IDs that are already intentionally random, like CreatePin() using Guid.NewGuid() inside the seed, we can simplify those to plain random GUIDs. But removing the seed globally would break existing internal ID relationships.

dirtkiller23: Forgot to clarify that an issue that occurs with determinstic - non/random GUIDs is when multiple .vnmgraph files are exported all at once. If they're opened in a editor, attempting to open ChildGraph/ReferencedGraph node will cause said editor to throw an error about similar IDs. Plus random GUIDs for me i think didn't cause an issue when diffchecking the newly compiled graph from a decompile against the original graph file but that could just not break with random GUIDs.

kristiker: Combine the resource file name in the input seed. But this will still be a problem if the user exports the resource twice for any reason. Just use Guid.NewGuid()

kristiker: >  If MakeGuid always returned a random GUID, those references could point to non-existent parameter nodes depending on traversal order.

This can be solved. Its better to use guids for what they are meant for.

STL1te: > This can be solved. Its better to use guids for what they are meant for.

I agree. Resolved that in last commit

PR #1157, opened 2026-05-15
Pulse graph viewer

LionDoge:
Adds a Pulse graph viewer for `vpulse` files. Pulse is a node based scripting system.

## What's left unhandled/untested
- Conditional loops - There are no official examples, besides ones from [my unofficial editor](https://github.com/LionDoge/vpulse-editor). I think this is something that could be implemented when there's an actual official graph that uses it, my editor might differ from the way the official one works. Besides, it also would leave more artifacts in the form of debug info that's included in the graph file (mine doesn't) which could help determine the loop structure consistently.
- Blackboards (or external graphs) - also no example of this used anywhere for now.

## What could be added or improved now
- Some static UI panel for displaying general info about graph, as well as the variables list, which currently is crudely placed as separate nodes.
- Layouting algorithm doesn't work too well for these types of graph, often making a bit of a mess, I would appreciate some help or pointers here, since I don't also want to break existing code, and I haven't learned how the existing graph code works.
- Text inside the nodes often overlaps or escapes the node area. I did not touch that either.
- Different node colors per the node type, and different font colors inside nodes depending on what information they represent.
- Refactoring the code - it's made to just work for now, not be pretty. There are still some repeated or unneeded sections.

## Examples
I laid out nodes here by myself, also it might differ slightly from the newest commits
- de_train_script (CS2)
de_train_script (CS2)

- Wingman script (CS2)
Wingman script (CS2)

- dl_midtown (Deadlock)
dl_midtown (Deadlock)

### DIFF
--- a/GUI/Types/Graphs/PulseGraphViewer.cs
+++ b/GUI/Types/Graphs/PulseGraphViewer.cs
@@ -3,6 +3,7 @@ using System.Drawing;
 using System.Linq;
 using System.Text;
 using System.Windows.Forms;
+using System.Xml.Linq;
 using GUI.Types.GLViewers;
 using GUI.Utils;
 using SkiaSharp;
@@ -106,6 +107,7 @@ internal class PulseGraphViewer : GLNodeGraphViewer
     private IReadOnlyList<KVObject> callInfos;
     private readonly Dictionary<int, Dictionary<int, SocketIn>> instructionInputActionSocketMap = [];
     private readonly List<NodeCallInfo> callNodesToResolve = [];
+    private readonly Dictionary<int, HashSet<(int, int)>> loopInstructionMap = [];
 
     struct NodeCallInfo
     {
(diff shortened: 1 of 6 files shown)


GUI/Types/Graphs/PulseGraphViewer.cs
```diff
@@ -0,0 +1,1240 @@
...
+                    {
+                        switch (cellType)
+                        {
+                            case CellType.PublicOutput:
+                                {
+                                    var outputIndex = cells[cellIdx].GetInt32Property("m_OutputIndex");
+                                    if (outputIndex == -1)
+                                        break;
+
+                                    var publicOutput = publicOutputs[cellIdx];
```
xPaw: is this supposed to be outputIndex

LionDoge: Yeah, that's a mistake, thanks.

xPaw:
Fix the build warnings please (you can build for release locally to get them).

p.s. join the discord

LionDoge:
The release build is stuck for me as well, just like the action, I don't know if this is something that I introduced or not. I'd rather not wait 3 hours with my PC going full turbine mode just for one compile. Disabling code analysis helps, but then I can't check for these warnings. I tried to look up why it could take that long, but I couldn't find anything.

I blindly tried to add 'using' like the warnings here suggested, but that just ends up crashing (cause these nodes should probably be disposed of when the graph ends rendering) not sure what's triggering these warnings anyways, since I don't see this being an issue in the Animgraph viewer. I don't usually program in C# so I've just assumed something like this should be handled by GC, but apparently not.

I'll try to figure it out layer, but I'll probably try working out some other things first

Poggicek:
The 3 hour build is caused by the `DisposeObjectsBeforeLosingScope` (CA2000) analyzer (the one that's actually throwing errors in the build rn) as reported by dotnet-stack against running csc. It's already disabled for debug builds https://github.com/ValveResourceFormat/ValveResourceFormat/blob/1e3813e0a994e34d22afd8e17b27615c93248ea6/Directory.Build.props#L43-L46

I got Claude to refactor all the closures in `CreateGraph`  into private methods and that instantly fixed the compile time.

image

Here's claude's patch of his iterative work trying to fix the compilation which you can use to tracking it further down
[claude.patch](https://github.com/user-attachments/files/27897906/claude.patch)

LionDoge:
Apart from the viewer logic itself, I had to touch some other general graph code to fix text not fitting the nodes for instance. There are probably better solutions for that, but that seemed to work good enough with the existing systems.

Every existing graphs in all games so far, so Deadlock, CS2, Dota2 can be opened fine, and seem to look correct. Apart from that, some community made graphs compiled with vpulse-editor should also work, but it's not a 100% guarantee, as they're missing some debug info at this point in time.

kristiker (approved):
👍🏼

Issue #866, opened 2024-11-22
YCoCg-R Conversion fails for SteamVR/Steamtours Assets marked as YCoCg


zahndy:
### Describe the bug

Results are quite wrong, see attached screenshots.

My guess is that valve did something different here but I can't figure out what.

SteamVR Home:
![Image](https://github.com/user-attachments/assets/b98ad2dc-111c-4166-b400-3a256142c45f) ![Image](https://github.com/user-attachments/assets/5663ed3c-4cd6-4cc1-ae7c-d69118e3cca8)
ingame:
![Image](https://github.com/user-attachments/assets/10b4bb55-8aaf-4666-a0a4-b5b6bb553175)

Result is the same when opening skyboxes from other environments from the workshop.

### Expected behavior

To view the colors that are displayed ingame.
(Steam\steamapps\common\SteamVR\tools\steamvr_environments\game\bin\win64\steamtours.exe +vr_enable_fake_vr_test)

### Relevant log output or exception with stacktrace

```shell

```

### What game does this happen in

SteamVR, steamtours

### Reference file in a Source 2 game

vpk:C:\Program Files (x86)\Steam\steamapps\common\SteamVR\tools\steamvr_environments\game\steamtours\pak02_dir.vpk:materials/skyboxes/vesper_peak_png_fbeb5735.vtex_c
vpk:C:\Program Files (x86)\Steam\steamapps\workshop\content\250820\1449922161\1449922161.vpk:materials/skyboxes/vesper_peak_evening_psd_9f19891a.vtex_c
vpk:C:\Program Files (x86)\Steam\steamapps\workshop\content\250820\1449922161\1449922161.vpk:materials/skyboxes/vesper_peak_night_psd_e11de8e4.vtex_c

### What version of VRF are you using?

10.2.0 1926645494d5d73c5acbfa93d0a0a05153aa560d on Microsoft Windows NT 10.0.19045.0 (GPU: NVIDIA GeForce GTX 1080 Ti/PCIe/SSE2, Driver: 4.6.0 NVIDIA 561.09)

xPaw:
There's nothing in the material or shader that would indicate "YCoCg-R", it just says YCoCg at most. So maybe the sky shader is just doing some kind of color conversion on top.

zahndy:
I wrote "YCoCg-R" because it looks like S2V uses the lifted version of YCoCg even though its named "YCoCg ", sorry for the confusion.

[...]

PR #1163, opened 2026-05-25
Fix YCoCg texture colors when exporting and previewing

w1tcherrr:
YCoCg textures come out desaturated with a green tint when exported to PNG or viewed in the texture inspector (#1127).

The Co/Cg/scale channels are stored in sRGB gamma space, and the engine samples these textures as sRGB, so they get linearized before the YCoCg matrix. The sky render was already fixed this way in #866 (`f3a943ec`, adding `SrgbRead(true)`), but the CPU export path and the texture-viewer shader still decode the raw gamma values directly.

I switched Decode_YCoCg from Color to Vector4 so the conversion happens in normalized float space instead of the byte-based Color, which would lead to losing precision.

Reproducible with `C:\Program Files (x86)\Steam\steamapps\common\SteamVR\tools\steamvr_environments\game\steamtours\pak02_dir\materials\skyboxes\vesper_peak_png_fbeb5735.vtex_c`

The sky shader is unchanged, since it already worked. If you want changes let me know.

Before:
broken

After:
fixed

### DIFF
--- a/Renderer/Shaders/texture_decode.frag.slang
+++ b/Renderer/Shaders/texture_decode.frag.slang
@@ -32,7 +32,6 @@ uniform vec4 g_vInputTextureSize;
 #define Dxt5nm_AlphaGreen (1 << 4)
 #define ColorSpace_Linear (1 << 5)
 #define ColorSpace_Srgb (1 << 6)
-#define YCoCg_Srgb (1 << 8)
 
 uniform int g_nSelectedMip;
 uniform int g_nSelectedDepth;
@@ -244,12 +243,13 @@ void main()
         vColor.rgb = PackToColor(vec3(normalXy.xy, derivedZ));
     }
 
+    if ((g_nDecodeFlags & ColorSpace_Srgb) != 0)
+    {
+        vColor.rgb = SrgbGammaToLinear(vColor.rgb);
+    }
+
     if ((g_nDecodeFlags & YCoCg_Conversion) != 0)
     {
-        if ((g_nDecodeFlags & YCoCg_Srgb) != 0)
-        {
-            vColor.rgb = SrgbGammaToLinear(vColor.rgb);
-        }
         vColor.rgb = DecodeYCoCg(vColor);
         vColor.a = 1.0;
     }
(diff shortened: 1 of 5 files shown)


kristiker:
Great changes, but this breaks YCoCg decode in newer engine versions like Dota2 and CS2.

Or maybe its a panorama and skybox material difference. We need to be able to detect the right decoder based only on Texture Compiler info found on the resource edit block.

Left: new, Right: old
image

`vpk:C:/Program Files (x86)/Steam/steamapps/common/Counter-Strike Global Offensive/game/csgo/pak01_dir.vpk:panorama/images/avatars/avatar_sub_01_psd.vtex_c`

w1tcherrr:
You're right, I incorrectly applied it to some textures I shouldn't have.

I instead now use the CUBE_TEXTURE property now that is on the SteamVR skyboxes for this linearization instead of doing it unconditionally. This works now for both skyboxes and normal textures like this panorama texture.

I also noticed some things in the texture viewer that I adressed:
- I made sure that the sRGB conversion stays when toggling YCoCg on/off in the texture viewer.
- The texture conversion changes did not always apply instantly upon pressing the checkbox, but only when hovering the mouse over the image. I changed it so it always applies instantly. At least it did not always work instantly, when opening the GUI to a specific material programmatically (using the CLI).
- When enabling / disabling the "Software Decode" checkbox it didn't remember the mode (Cubic / Equirectangular) the user used before. I changed it so disabling the option restores the mode the user used before.

I also wanted to keep the YCoCg as one checkbox, which is why this logic with the hidden YCoCgSrgb flag is needed. If you want I can split it into two checkboxes, one for YCoCg and one for YCoCgSrgb each.

Renderer/Shaders/texture_decode.frag.slang
```diff
@@ -245,6 +246,10 @@ void main()
 
     if ((g_nDecodeFlags & YCoCg_Conversion) != 0)
     {
+        if ((g_nDecodeFlags & YCoCg_Srgb) != 0)
+        {
+            vColor.rgb = SrgbGammaToLinear(vColor.rgb);
+        }
```
kristiker: We should probably move to-linear color space conversion earlier. And the need for `YCoCg_Srgb` goes away, by setting `YCoCg_Conversion` and `ColorSpace_Srgb` together. We also get rid of the texture viewer checkbox visibility hacks.

w1tcherrr:
You're right, I didn't use this, because I didn't see it used anywhere else before (the ColorSpace_Srgb being toggled on by default), so I didn't really consider it.

This should still work, but now the ColorSpaceSrgb checkbox is toggled as it should be for the skybox vtex_c. Also the export (Save to disk) properly works. 

To be honest though, I not sure if the changes I made are what you meant by your comment.

kristiker (approved):
Yeah looks good!

PR #1164, opened 2026-05-25
Allow trailing bytes in file-size validation (fix #1119)

w1tcherrr:
Fixes #1119 by relaxing texture size validation to allow trailing bytes for all textures while still rejecting truncated textures. This should be safe, because the texture decoders only read the payload length, so any trailing bytes can be ignored. I think weakening the validation to remove such edge cases is worth it.

The map from the ticket does not reproduce the issue anymore, but I could reproduce it using this map: [Workshop Link](https://steamcommunity.com/sharedfiles/filedetails/?id=3369117065)

### DIFF
--- a/ValveResourceFormat/Resource/Resource.cs
+++ b/ValveResourceFormat/Resource/Resource.cs
@@ -318,23 +318,11 @@ namespace ValveResourceFormat
             {
                 if (ResourceType == ResourceType.Texture)
                 {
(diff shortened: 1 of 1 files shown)


xPaw:
I'm not convinced we need to do this, it has helped catch bugs in the past, and this only fails for a few files in workshop maps.

Issue #535, opened 2023-04-02
LODs are not exported to GTLF


xGitHubUser999x:
### Describe the bug

Starting with version 0.3.0.1412 the VRF stopped exporting LODs to GLTF. Whereas in version 0.2.2.1057 everything is exported properly.

### What Valve game does this happen in

Dota 2

### Reference file in a Source 2 game

models/items/axe/axe_blackthorn_head/axe_blackthorn_head.vmdl

### Expected behavior

The exported GLTF must contain both axe_blackthorn_head and axe_blackthorn_head_lod1 models.

### What version of VRF are you using? On what platform?

VRF: 0.3.1.1553; Platform: Windows 8.1 x64

xPaw:
This is intentional. Do you have a good reason to export lods?

PR #1182, opened 2026-06-09
Lod Level Selection (Model Viewer)

### LINKED ISSUE #657: Support viewing different level-of-detail in model viewer
We currently ignore everything except for lod0. If we figure out how to switch them on the fly for a single model, we can later use the same logic in the map viewer.

### LINKED ISSUE #1145: Add LOD level selection in the model viewer

I found that if some creators do not configure LOD0 correctly during model creation (for example, leaving it without any meshes), the Viewer is unable to preview the model. 
However, the game treats this as valid behavior, since it automatically skips empty LOD and to the next available LOD for rendering.

Viewer preview
Image

Workshop tool preview
> Default LOD
> Image

> LOD 0
> Image

> LOD 1
> Image

_No response_

_No response_

w1tcherrr:
## Add LOD level selection to the model viewer

I implemented runtime LOD selection in the model viewer and reworked how LOD data is handled. Empty-LOD0 models now resolve to the lowest populated level everywhere (viewer, glTF export, `SceneAggregate`) rather than rendering/exporting nothing. The lowest populated level is the new default instead of always LOD0.

I tested manually using multiple models, also doing some "round-trips" (`vmdl` -> resourcecompiler -> `vmdl_c` -> source2viewer -> `vmdl`). I attached these models ([lod_models.zip](https://github.com/user-attachments/files/28763699/lod_models.zip)):
- king: Single mesh flagged for all 8 LOD levels with no switch distances (mask `255`). The mask suggests "always shown, no real LOD", where the selector stays hidden and the decompiler does not emit a `LODGroupAll`.
- flatnose_truck: Real multi-level LODs plus a `LODGroupAll` mesh; the round-trip case where per-level `LODGroup`s and `LODGroupAll` decompile and recompile correctly.
- lod_test: Synthetic 5-LOD fixture (masks `1,2,4,8,16`, switches `0,5,10,15,20`); simple per-level geometry that verifies the automatic LOD switching easily.
- ctm_sas: Empty LOD0 with all meshes in LOD1; the original bug case (#1145) where the viewer showed nothing and the selector was wrongly hidden. Now it defaults to LOD1 instead of LOD0.

### Viewer
- New **Level of Detail** dropdown: `Auto` (mirroring Source 2's screen-size metric) plus a manual entry per level, where each level shows the metric range it covers.
- The selector only appears when switching levels actually changes the rendered geometry, so single-mesh "always shown" models (mask `0xFF`) don't get a useless dropdown. Empty levels are listed (and selectable) so you can still manually inspect them (even if they're empty).
- I show the active level in the viewport overlay: a line in the bottom-left corner above the FPS counter (`LOD: Auto (2)` while it is auto-resolving, `LOD: 2` when forced), which is hidden while the viewer is paused. The same line also shows in the selected-mesh stats overlay, which refreshes live on a LOD change.

### ModelLodInfo type
- Consolidated all the LOD mask/level bit math and level selection into a single `ModelLodInfo` type owned by `Model`, instead of reading `m_refLODGroupMasks` / `m_lodGroupSwitchDistances` raw in multiple places.

### Decompiler
- `vmdl` decompile re-emits a `LODGroupList` (one `LODGroup` per switch level) and preserves `LODGroupAll` for meshes present in every level, so a recompile reproduces the original LOD structure. Models with no switch distances stay LOD-free. If a mesh was authored in every level via separate `LODGroup` entries rather than a `LODGroupAll`, it still comes back as a `LODGroupAll` on decompile, since the compiler keeps only the per-level mask, so that distinction is lost.

### Notes
- Added `ModelLodInfo` unit tests and a `LODGroupList` decompile test.
- Closes #1145
- glTF export still emits a single LOD (the lowest populated one). Multi-LOD export via `MSFT_lod` could be added (requested in #535)
- I only tested using CS2 so far. I don't have any knowledge over other Source 2 games, since I never touched any of them (not even played).
- The selection logic sits in the `ModelSceneNode` and the `SceneAggregate`, so it also works when a model is loaded as part of a vmap. The dropdown and overlay are only in the model-viewer though. The vmap does not have such a mesh-stats overlay, so it probably makes no sense to display it there.


Renderer/Renderer/SceneAggregate.cs
```diff
@@ -172,7 +181,7 @@ private IEnumerable<Fragment> CreateFragments(KVObject aggregateSceneObject)
                 var flags = fragmentData.GetEnumValue<ObjectTypeFlags>("m_objectFlags", normalize: true);
                 var lodGroupMask = fragmentData.GetUInt32Property("m_nLODGroupMask");
 
-                if (lodGroupMask > 1)
+                if (lodGroupMask != 0 && (lodGroupMask & lowestLodBit) == 0)
```
kristiker: This isn't clear. I'm guessing it allows lod1 or 2 to pass in case zero was empty. Needs to be commented.

PR #1184, opened 2026-06-11
Track navigation history in package viewer

### LINKED ISSUE #1112: Track navigation history in package viewer

add <> buttons here for going forward and back

Image

_No response_

_No response_

w1tcherrr:
Buttons

(closes #1112)

Adds back/forward navigation to the package viewer, like a file browser. You can retrace the folders you've visited via two toolbar buttons or the mouse side buttons (XButton1/XButton2).

Behavior
- Every folder shown in the list view is recorded as a history entry. There are two stacks (one forward / one backwards).
- Searches are recorded as history entries using this NavigationEntry class to record which kind of navigation it is (normal navigation vs searches).
- Buttons are disabled when there's nowhere to go in that direction.
- suppressHistoryRecording flag makes sure the back/forward navigation does not re-record entries when "replaying" the history.

Choices
- New NavigateBack/NavigateForward icons are the existing SortUp/SortDown glyphs rotated 90°. I still made them into new files so they can be replaced independently of the SortUp/SortDown glyphs.
- Now the ThemedButton does not draw its image identically whether enabled or disabled. Since these buttons start disabled, it now dims the disabled image via a ColorMatrix (alpha 0.35) while keeping the same scaling. Could not use ControlPaint.DrawImageDisabled because it ignores the target rectangle and renders the icon at the wrong size. This was not used by other buttons before (no other ThemedButton with an image is ever disabled), so it does not change any existing behaviour.

Let me know about any requested changes / issues :)


xPaw:
Seems fine overall, there might be some bug with vpk editing where you delete a folder.

1. Add tooltip text to the buttons in designer
2. I think these x buttons need to be ignored in mousedown (or it should filter to left/right only).

w1tcherrr:
Good points. I didn't think about the vpk editing feature, because I've never used it (since it's not available for pre-existing vpks). Now the history prunes all entries that are either the deleted node or any of its sub-nodes.

Searches still stay in the history on purpose - replaying one re-runs it against the current tree, so deleted files will just drop out of the results instead of showing up again.

I added the ToolTips and added the ignore in mousedown.

The mouse side buttons are no longer handled via a MouseUp handler on the list view. Instead the control overrides WndProc and handles WM_APPCOMMAND (BROWSER_BACKWARD/FORWARD), same pattern as MainForm.Frame.cs. This makes more sense, so the back/forward mouse buttons work also in the other views (tree view, toolbar, splitter). Otherwise it felt unintuitive that the mouse buttons would not work for navigation when hovering e.g. in the tree view.

I replaced the arrows with the images you linked (https://lucide.dev/icons/arrow-left and https://lucide.dev/icons/arrow-right).

kristiker:
Tested this, working good!

image

Were these colors intended on light theme?

w1tcherrr:
Good point. I made a light mode version for the SVG and also made the arrows a bit longer (in my opinion the head is a bit unproportionally big compared to the shaft) and a bit thicker.

I think it didn't look terrible before, so I didn't notice it, but I think this is a bit nicer.

dark

---

light

PR #1193, opened 2026-06-19
Export animation graph clips & bake transforms for gltf

### LINKED ISSUE #994: An ability to choose what skeletal animation to export (AnimGraph2)

With CS2 being updated to the AnimGraph2, I discovered that VRF only exports a single skeletal animation (.vnmclip_c), ignoring all the others. Since all weapon animations in CS2 are now split into hand animations and weapon animations, it is not possible to export weapon animations.

Export each skeletal animation to a different file (e.g, "my_file_viewmodel.vnmskel", "my_file_ak47.vnmskel") or An ability to override skeletal animation to export.

_No response_

### LINKED ISSUE #1135: Robot Repair: Animated models GlTF export with incorrect setup
### Describe the bug

When exporting some animated models from Robot Repair like `aperture/models/tape_cabinet.vmdl_c`, as GlTF it seems to have some overblown scale applied to it. Even after CTRL + A applying all transforms of both the armature and the mesh, some bones stretch wildly everywhere, namely this model's `loop` animation has the tape arms stretch five meters out of the model on both sides.

### Expected behavior

Models decompiled and exported should have the correct scale and no animation errors.

```shell

```

Robot Repair (The Lab)

`aperture/models/tape_cabinet.vmdl_c`

### What version of S2V are you using?

GitHub release 19.1

### Have you tried the latest dev build?

Yes

### LINKED ISSUE #1174: When exporting models to gltf format, animations are not saved with them as before.
### Describe the bug

When exporting a model (for example, a SAS KT model) to gltf format, 2,000 animations that work and display in the program itself are not exported. Please fix it!

Image

### Expected behavior

When exporting a model to gltf format, animations should be in the data base as before.

```shell

```

Counter-Strike 2

agents/models/ctm_sas/

### What version of S2V are you using?

19.2.6343 c72208352 on Microsoft Windows 10.0.26200 (X64) (GPU: Radeon (TM) RX 480 Graphics, Driver: 4.6.0 Core Profile Context 26.1.1.251211)

### Have you tried the latest dev build?

Yes

w1tcherrr:
## 1 - Animation graph clips (fixes #1174)

Since ag2 models reference their animations as separate nmclip (.vnmclip) files through an animation graph (instead of embedding them) they were not written out when decompiling the model, so a decompiled AnimGraph2 model lost most of its animations (except for the few embedded ones).

A new helper, `AnimationGraphLoader.cs`, walks the model's animation graph(s), following m_animGraph2Refs and recursing into nested graphs, returning the referenced clip names, de-duplicated.

Both exporters consume it: glTF writes one animation track per clip into the single .glb (retargeted onto the model's joints by bone name), and DMX writes each clip as its own file alongside the model.

## 2 - Conversion baking (fixes #1135)

This wasn't only a Robot Repair issue (where it was reported) - it broke every skinned model in every game.

The source-to-glTF axis swap and inch-to-meter scale used to live on the armature/mesh node transform, so skinned models broke (exploded) when applying transforms in Blender. The conversion is now fully baked into the geometry and animation data instead, so the armature stays identity-scaled. The conversion/baking code is consolidated in `GltfModelExporter.Conversion.cs`.

It basically is the same as before, but now it doesn't break in blender when applying transforms.

## 3 - CLI option --gltf_animation_list

This option already existed and it kind of fixes #994 - though the export-everything from 1.) is what actually solves it, this is just the "pick specific ones" part, and it's CLI only, not possible in the GUI.

It also didn't really work for Ag2 before, so I fixed a few things:

- it now matches clips by their leaf name (e.g. "idle_knife"), not just the full resource path - before, only the full path matched, so for Ag2 clips you couldn't filter easily.
- it trims whitespace and warns when names in the list don't match anything.

And obviously you can also just export the model with all its animations and delete the ones you don't want.

The ticket mentions vnmskel but I think they must have meant vnmclip.

### DIFF
--- a/CLI/Decompiler.cs
+++ b/CLI/Decompiler.cs
@@ -1422,7 +1422,7 @@ namespace CLI
             return highestByShader.Values.Select(x => x.FilePath).ToHashSet();
         }
 
-        private static void DumpContentFile(string path, ContentFile contentFile, bool dumpSubFiles = true)
+        private void DumpContentFile(string path, ContentFile contentFile, bool dumpSubFiles = true)
         {
             if (contentFile.Data != null)
             {
@@ -1431,7 +1431,10 @@ namespace CLI
 
             foreach (var additionalFile in contentFile.AdditionalFiles)
             {
-                DumpContentFile(Path.Combine(Path.GetDirectoryName(path)!, Path.GetFileName(additionalFile.FileName)), additionalFile);
+                var additionalPath = additionalFile.KeepFullPath && OutputFile != null && (IsInputFolder || Directory.Exists(OutputFile))
+                    ? Path.Combine(OutputFile, additionalFile.FileName)
+                    : Path.Combine(Path.GetDirectoryName(path)!, Path.GetFileName(additionalFile.FileName));
+                DumpContentFile(additionalPath, additionalFile);
             }
 
             if (dumpSubFiles)
--- a/CLI/Decompiler.cs
+++ b/CLI/Decompiler.cs
@@ -1431,6 +1431,10 @@ namespace CLI
 
             foreach (var additionalFile in contentFile.AdditionalFiles)
             {
+                // Additional files (animation-graph clips) carry their full resource path. With a real output
+                // directory we keep it; otherwise we flatten to the leaf name next to the parent file, which can
+                // collide on shared names. Resolving these relative to the parent's output path properly needs a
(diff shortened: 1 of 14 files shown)


kristiker:
The gltf transform baking will break some users workflow, but perhaps its worth it. Could make it optional later through a cli arg. Otherwise looking good!

PR #1194, opened 2026-06-21
Extract point_template child entities and resolve attachment parenting

### LINKED ISSUE #958: Correctly decompile `point_template` child entity lumps in vmap
Currently the child lumps are just decompiled as-is, which does not place them correctly in world. Then `point_template` also needs to correctly link to these entities with `TemplateXX` keys.

See 46e515a67a574a6bed61f6ab72f07c8837e3da93 for the same fix in gltf.

w1tcherrr:
## 1 - point_template

When the compiler bakes a `point_template`, the templated entities get pulled into separate entity lumps, stored relative to the template's origin and rotation (scale is not propagated!). The decompiler already emitted these children, but at those raw template-relative coordinates - so they landed in the wrong place. Each child's stored transform is now composed with its template's rigid transform to put it back where it was authored - confirmed by a compile -> decompile round-trip.

The compiler also clones entities across several templates (when the same entity is referenced by more than one). These clones keep the original `hammeruniqueid`, so they're folded back into one entity - otherwise each recompile would multiply the spawns.

The traversal the renderer and glTF exporter each did on their own is now one shared helper (`EntityLumpTraversal.cs`); `MapExtract` keeps its own walk since it additionally needs orphan-lump emission and parent-transform threading. Added `Tests/Files/point_template_test.vpk` + `PointTemplateExtractTest.cs`.

Closes #958

In the map viewer, each point_template and the entities it spawns are grouped into a dedicated "Template Entities" world layer, so they get their own checkbox in the layer selector and can be shown or hidden together instead of being scattered across the regular entity/tool layers. This means that point_template entities are not part of normal "Tool Entities", contrary to the current design. Otherwise the entities the point_template spawns would be visible even when the point_template itself is not.

The lines from each point_template (when selected) to its entities - like in Hammer - are not drawn currently.

## 2 - attachment parenting (map viewer)

Map props can be attachment-parented (`parentname` + `parentattachmentname`) - glued onto an attachment point on another model. The engine snaps them onto the attachment point at load, but VRF drew them at their stored Hammer position, so they ended up in the wrong spot/rotation - most visibly de_train's cargo cars (180° off).

Similarly to point_template the scale is not propagated. So they're snapped onto the parent's attachment after load - the attachment's position and rotation with the child's own scale, never the parent's. Model-less entities (drawn as default-entity icons) only follow the attachment's position. This happens regardless of `uselocaloffset` (the engine ignores it for attachment parenting too). Decompiling instead keeps the stored Hammer positions.

Barely used in official maps - de_train's cargo train is the only place I could find where it visibly matters. There are some other places, but usually entities like ambient_generic are snapped to radios or similar (so the sound is emitted in the right place).

## Notes

The whole parenting / snapping thing is a bit hacky. VRF would need a proper parenting hierarchy for all entities to resolve this cleanly. 

This map can be used to test the attachment-parenting (ingame vs S2V): [maps.zip](https://github.com/user-attachments/files/29169315/maps.zip). Also the train-carts on train are correct now.

The point template logic is tested thoroughly in `PointTemplateExtractTest.cs` and can be tested using the vpk in the testing suite.


Renderer/Renderer/World/WorldLoader.cs
```diff
@@ -228,6 +235,57 @@ public void LoadEntities()
...
+            {
+                return false;
+            }
+
+            attachment = parentNode.GetAttachmentTransform(attachmentName);
+            return true;
+        }
+
+        private static bool IsAttachmentParented(Entity entity)
+            => !string.IsNullOrEmpty(entity.GetStringProperty("parentname"))
```
xPaw: Is ContainsKey not enough here, do they store empty strings?

Renderer/Renderer/World/WorldLoader.cs
```diff
@@ -1094,6 +1126,12 @@ void LoadEntity(string classname, Entity entity)
 
                 modelNode.Flags |= entityFlags;
 
+                var modelTargetName = entity.GetStringProperty("targetname");
+                if (!string.IsNullOrEmpty(modelTargetName))
+                {
+                    entityModelNodesByName[modelTargetName] = modelNode;
```
xPaw: How many lookups realistically need to be done, can't FindEntityByKeyValue be used?

ValveResourceFormat/Utils/EntityLumpTraversal.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <c>point_template</c> entities reference. Template children inherit the template's rigid transform (no scale).
+        /// </summary>
+        /// <param name="lump">The root entity lump.</param>
+        /// <param name="resolveChildLump">Resolves a referenced child lump by name, or <see langword="null"/> if it can't be loaded.</param>
+        /// <param name="rootTransform">Transform applied to top-level entities.</param>
+        /// <param name="onMissingChildLump">Called with the lump name when a referenced child lump can't be resolved.</param>
+        /// <returns>Each entity with its parent transform.</returns>
+        public static IEnumerable<TraversedEntity> EnumerateEntities(
+            VEntityLump lump,
+            Func<string, VEntityLump?> resolveChildLump,
```
xPaw: We typically pass in the file loader, which should work here too.

xPaw:
> VRF would need a proper parenting hierarchy for all entities to resolve this cleanly.

Whats missing and what would it take?

kristiker:
We can add attachments to ModelSceneNode. The model updates children transforms on `Update`. I don't like setting transform just once on WorldLoader, because if the parent model moves (in the future via scripts or user gizmo) they will stay in place.

Renderer/Renderer/World/WorldLoader.cs
w1tcherrr: FindEntityByKeyValue returns the Entity but I need the ModelSceneNode. I replaced it with FindModelNodeByTargetName and removed the dictionary & list

Renderer/Renderer/World/WorldLoader.cs
w1tcherrr: I replaced it with a null check

w1tcherrr:
@kristiker you're right about not setting it once, I switched to the new `ModelSceneNode` attachment API, so children get updated instead of having their transform snapped a single time in `WorldLoader`. 

The child keeps its own scale, and we must not propagate the parent's scale (Source 2 doesn't, probably due to shearing concerns). I've fixed that.

@xPaw what's missing for a proper parenting hierarchy?

The ModelSceneNode update is great already, but only a ModelSceneNode can be a parent. Theoretically, something like env_sky (SpriteSceneNode) can also be a parent. 

What the usecase for that is I don't know, but perhaps something like info_target could have usecases for this (e.g. snapping an object to a target node?). 

I'm not sure how much more to parenting there really is. The ModelSceneNode is also used for rendering obviously, so in my head it would have made sense if the Entities themselves already supported a parenting hierarchy.

kristiker (approved):

kristiker:
Nice! Base entity (SceneNode) parenting could be added in the future if there's usecases.

PR #1192, opened 2026-06-19
Add native Linux Source 2 Viewer GUI

dede177:
## Summary

This introduces a Linux-native Source 2 Viewer application while keeping the existing Windows WinForms GUI unchanged.

The current branch provides a usable Linux-first GUI for opening VPK archives and loose Source 2 resources, browsing and filtering package contents, and rendering models and maps with the existing VRF renderer. It also includes renderer fixes found while exercising assets through the Linux OpenGL path.

This is submitted as a draft because the native viewer is functional, but it does not yet have full feature parity with the Windows GUI.

## Technology

- **.NET 10 / C#** for the application and shared code.
- **Avalonia 12** for the native cross-platform desktop shell, controls, file pickers, input, drag-and-drop, and theming.
- **Avalonia `OpenGlControlBase`** for the viewport lifecycle and OpenGL surface.
- The existing **Renderer** project and **OpenTK/OpenGL** rendering pipeline; this does not introduce a second renderer or a web/Electron layer.
- Existing **ValveResourceFormat** and **SteamDatabase.ValvePak** loaders for Source 2 resources and VPK archives.

## Current scope

- Adds `Source2Viewer.App`, targeting `net10.0`.
- Opens VPK archives and loose compiled Source 2 resources.
- Provides package navigation, search/filter/sort, preview, extraction/decompile actions, drag-and-drop, and recent-file startup behavior.
- Renders models, worlds/maps, particles, navigation resources, and sky content through the shared renderer.
- Supports orbit/pan/look navigation, wheel zoom, and model/entity focus.
- Exposes map cameras, layers, physics groups, entities, model groups, materials, LODs, hitboxes, skeletons, animations, render modes, and renderer debug toggles where available.
- Keeps the existing Windows GUI project intact.
- Fixes OpenGL buffer offsets/VAO state, sampler initialization, and missing vertex-color fallback issues observed on the Linux path.

## Plan

1. **Review and stabilize this Linux foundation**
   - Validate more games, maps, weapons, props, and GPU/driver combinations.
   - Address renderer regressions with small reproducible assets and focused tests where practical.
   - Keep Windows GUI behavior/builds unchanged.

2. **Close viewer parity gaps in follow-up PRs**
   - Add the remaining resource-specific viewers and richer document tabs.
   - Match package browsing, export, properties, audio/image, and navigation workflows from the Windows application.
   - Split follow-up work into reviewable viewer-specific changes instead of expanding this PR indefinitely.

3. **Harden the platform boundary**
[...]

PR #1195, opened 2026-06-21
Renderer fixes

dede177:
This pulls the renderer-only changes out of #1192. There are no Linux UI changes in this branch.

The changes are:

- Apply `m_nBindOffsetBytes` when binding vertex buffers and calculating the index start offset.
- Include the complete vertex-buffer binding in the VAO cache key so draw calls using different offsets do not reuse the wrong VAO.
- Use a constant white value when a shader expects `vCOLOR` but the mesh has no color stream. The previous one-element buffer advanced once per vertex.
- Initialize active sampler uniforms to their reserved or material texture units. This avoids different sampler types all starting on texture unit 0 before their textures are bound.
- Treat all OpenGL sampler types as textures, including arrays, cube, shadow, and integer samplers.

These fixed blank or incomplete models and map entities seen while loading CS2 assets. The buffer-offset changes also replace the zero-offset debug assertions currently on `master`.

Tested with:

- `dotnet format Renderer/Renderer.csproj --verify-no-changes --no-restore`
- `dotnet build Renderer/Renderer.csproj --configuration Release --no-restore`
- `dotnet test Tests/Tests.csproj --configuration Release --no-restore` (275 passed, 2 skipped because local sample game data was unavailable)

### DIFF
--- a/Renderer/Renderer/GPUMeshBufferCache.cs
+++ b/Renderer/Renderer/GPUMeshBufferCache.cs
@@ -19,7 +19,7 @@ namespace ValveResourceFormat.Renderer
         private readonly Dictionary<string, GPUMeshBuffers> gpuBuffers = [];
         private readonly Dictionary<VAOKey, int> vertexArrayObjects = [];
 
-        private record struct VAOKey(string MeshName, int Shader, int VertexIndex, int IndexIndex);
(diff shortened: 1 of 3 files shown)


kristiker:
The assert was not being hit on any file, so that fixes nothing.

PR #1197, opened 2026-06-21
Apply scene object transform to decompiled hammer meshes

### LINKED ISSUE #895: Some stuff appears at 0/0/0 after decompiling de_train to vmap
### Describe the bug

Hello, after decompiling the tree model (screen 1), it turned out to be at 0 coordinates + grass that should not have collision has it
Image

Image

### Expected behavior

Tree out of place + grass without collision

```shell

```

CS2

de_train

### Have you tried the latest dev build?

Yes

### What version of VRF are you using?

11.1.4116 5198d7f3a4f058a11b4feaf83e8f3fe0202b87c6 on Microsoft Windows NT 10.0.19045.0 (GPU: NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2, Driver: 4.6.0 NVIDIA 572.47)

w1tcherrr:
Fixes #895

Scene objects that get converted to a Hammer mesh (instead of a prop_static) were exported without their transform. Not just the scale/orientation, the whole transform was dropped, so these meshes ended up at 0, 0, 0. Most visible on de_train. The renderer placed them correctly from the scene object's m_vTransform already, only the map export ignored it.

Pretty rare in practice: across all official CS2 maps only 50 baked hammer-meshes have a non-identity transform (all others have an identity transform). 28 of those are on train, 10 on alpine, 10 on warden and 2 on italy. Every other official map is unaffected.

### DIFF
--- a/ValveResourceFormat/IO/HammerMeshBuilder.cs
+++ b/ValveResourceFormat/IO/HammerMeshBuilder.cs
@@ -647,7 +647,7 @@ namespace ValveResourceFormat.IO
             return mesh;
         }
 
-        public void AddVertices(VertexStreams streams, Vector3 positionOffset)
+        public void AddVertices(VertexStreams streams, Vector3 positionOffset = new Vector3())
         {
             vertexStreams = streams;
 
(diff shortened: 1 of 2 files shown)


w1tcherrr:
Original bug (after export to hammer):
trees_hammer_broken

After fix (in hammer):
trees_hammer

S2V comparison (this already worked, but shows that hammer looks correct now)
trees_s2v

PR #1202, opened 2026-06-23
Increase svg to png export resolution

### LINKED ISSUE #982: Export svg with anti-aliasing

I was exporting vsvg_c files from Deadlock as png and I noticed there was some noticeable aliasing in the images. If I export the same image as svg, open them with Photoshop's anti-aliasing and export as png, I can get the image in the same resolution with better quality and smaller file size. See comparison:

Image

Image

Apply anti-aliasing natively when using the option to decompile vsvg files as png

_No response_

w1tcherrr:
Fixes #982

When exporting vsvg files to png they came out aliased. The cause isn't missing anti-aliasing though, because Svg.Skia already applies AA by default (https://github.com/wieslawsoltes/Svg.Skia/blob/v5.0.0/src/Svg.Model/Services/PaintingService.cs#L640). So the "aliasing" is just a symptom of low resolution on the export. That's why, instead of saving the SVG at exactly the size it's rendered at in the viewer, I floored the export at 1024 (if you export an SVG as PNG I don't think we have to be economical with the resolution, since that conversion is inefficient in itself anyway - files are still only a few kB).

Also the DrawPicture call had a no-op IsAntialias = true. When a paint is passed to DrawPicture this flag is not used.

Exports now rasterize at a resolution derived from the native size scaled by the current zoom in the viewer, rounded up to the nearest power of two and clamped to 1024 - 4096. So effectively there are three export resolutions now: 1024, 2048 or 4096.

In the future it would make most sense for the user to select it manually (instead of it being derived from the zoom level), but I think that makes more sense as part of the GUI rework. 

I also tried some stuff like supersampling, etc. but everything looks worse than what Skia does by default.

### DIFF
--- a/GUI/Forms/SvgExportForm.cs
+++ b/GUI/Forms/SvgExportForm.cs
@@ -236,5 +236,20 @@ namespace GUI.Forms
             var scale = longEdge / SourceLongEdge;
             return ((int)(sourceWidth * scale), (int)(sourceHeight * scale));
         }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                formatComboBox.Dispose();
+                presetComboBox.Dispose();
+                customNumeric.Dispose();
+                resolutionLabel.Dispose();
+                customRow.Dispose();
+                outputLabel.Dispose();
+            }
+
+            base.Dispose(disposing);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


xPaw:
Feels kinda weird to always export 16px icons as 1024px. What if it simply was 2x or 4x?

w1tcherrr:
I added a popup, that let's you select between SVG, PNG, JPG. When selecting PNG or JPG you can pick a resolution from the presets or choose a custom size (can't go below native). Aspect ratio is preserved.  I know it isn't the prettiest, but I think it's good enough until a new UI is made.

svg
png
sel
num

PR #1203, opened 2026-06-23
Select correct range on shift-click in the package list

w1tcherrr:
The native selection in the list view does not work properly. When shift-click selecting from e.g. `ctm_fbi_variantf (directory)` to `ctm_fbi_variantf (vmdl)` it incorrectly select rows to the left of the directory in the upper row and to the right of the bottom row.

Default behaviour:
incorrect

Correct behaviour:
correct

### DIFF
--- a/GUI/Types/PackageViewer/BetterListView.cs
+++ b/GUI/Types/PackageViewer/BetterListView.cs
@@ -30,6 +30,18 @@ namespace GUI.Types.PackageViewer
 
         protected override void WndProc(ref Message m)
         {
+            const int WM_LBUTTONDOWN = 0x0201;
+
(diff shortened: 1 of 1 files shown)


xPaw:
Is this a bug in the winforms control?

w1tcherrr:
It's just the native behaviour, in icon/grid view it selects everything inside a bounding box, whereas the list/details views use a plain index range. Windows Explorer overrides this in its own view too. See [Wine](https://github.com/wine-mirror/wine/blob/05421efde8f3e2dc6ffb1883ac1fa282839224e1/dlls/comctl32/listview.c#L3649-L3688)

PR #1208, opened 2026-06-27
Mark vcs 71 as supported

kristiker:
No meaningful change in this PR. Scanned all dota 2 shaders, all parse sucessfully with the CLI stats feature.

If there is a breaking change it will show up in other games when they also update their engine version.


xPaw:
I think 70 was the same, with no code changes.

PR #1213, opened 2026-07-03
Particle renderer fixes

kristiker:
**Crashes / NaN**
- `PointList` path interpolation indexed out of bounds (open and closed paths)
- NaN poisoning from zero-vector normalization in `Particle.Normal`, `RenderTrails`, `NormalizeVector`, `MaxVelocity` and `RingWave` (0 particles per orbit)
- `NoiseEmitter` threw on negative `m_flOutputMax`; random index overflow in `RandomBetweenPerComponent` call sites

**Wrong behavior**
- `CurveFitting`: segment evaluation used absolute x with segment-relative coefficients, and the last curve segment was never interpolated
- Culling: particle AABB subtracted `m_BoundingBoxMin` instead of adding it, producing boxes that didn't contain their particles
- Rotation units unified to radians (constants parsing, `Spin`/`SpinYaw`, `SpinUpdate`) — spins were 57x too fast or slow depending on the writer
- `PositionLock`: per-frame transform delta was computed per particle (only the first particle followed the CP), particles snapped onto the CP instead of translating with it, and the default time window made it a no-op
- `RandomYawFlip` probability was inverted; `RemapSpeedToScalar` remapped particle ID instead of speed
- `ContinuousEmitter` burst all pre-start time on its first emitting frame; `NoiseEmitter` kept emission credit across restarts
- `RenderTrails`: screen-aligned trails rendered at the world origin, length fade-in compared normalized age against seconds, quads were back-face culled, wrong UV offset without a spritesheet
- Light renderers divided already-normalized colors by 255 (near-black lights); literal colors now normalized once in the provider
- `AttributeMapping`: input/output ranges were sorted independently, reversing inverting remaps; `RemapBiased` never clamped in its default input mode
- `MathUtils.Fract` was sign-broken for negative inputs (noise discontinuities); `Noise.PseudoRandom` degenerated at large times
- `SetControlPointRotation` never accumulated (frame-rate-dependent static orientation) and passed degrees as radians
- `RandomUniformOffsetVectorProvider` hid the base method with `new`, so the offset was silently ignored

### DIFF
--- a/Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
+++ b/Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
@@ -38,7 +38,8 @@ namespace ValveResourceFormat.Renderer.Particles.Emitters
             this.particleEmitCallback = particleEmitCallback;
 
             time = 0f;
-            particlesToEmit = 0f;
+            // Source 2 starts the accumulator at 1 so short or slow emitters (rate * duration <= 1) still emit
+            particlesToEmit = 1f;
 
             IsFinished = false;
         }
@@ -63,15 +64,15 @@ namespace ValveResourceFormat.Renderer.Particles.Emitters
 
             if (time >= nextStartTime && (nextEmissionDuration == 0f || time <= nextStartTime + nextEmissionDuration))
             {
-                // Calculate current emission rate based on noise
-                var noise = Noise.Simplex1D(time * noiseScale.NextNumber(particleSystemState));
+                // Calculate current emission rate based on noise, remapped from [-1, 1] to [0, 1]
+                var noise = (Noise.Simplex1D(time * noiseScale.NextNumber(particleSystemState)) * 0.5f) + 0.5f;
                 var emissionMinValue = emissionMin.NextNumber(particleSystemState);
                 var emissionMaxValue = emissionMax.NextNumber(particleSystemState);
                 var emissionRate = emissionMinValue + noise * (emissionMaxValue - emissionMinValue);
 
-                // Aggregate emission into num of particles to emit.
-                // Min/Max instead of Clamp because a negative m_flOutputMax must not throw.
-                particlesToEmit += Math.Max(0f, Math.Min(emissionRate, emissionMaxValue)) * frameTime; // Limit the amount of particles to emit at once in case of refocus
+                // Aggregate emission into num of particles to emit. The noise remap already targets
+                // the min/max range, so no upper clamp (an inverted range must not kill the emitter).
+                particlesToEmit += Math.Max(0f, emissionRate) * frameTime;
 
                 // If nr of particles to emit is > 0, emit it
                 while (particlesToEmit > 1.0f)
--- a/Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
+++ b/Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
@@ -38,6 +38,7 @@ namespace ValveResourceFormat.Renderer.Particles.Emitters
             this.particleEmitCallback = particleEmitCallback;
 
             time = 0f;
+            particlesToEmit = 0f;
 
             IsFinished = false;
         }
@@ -68,8 +69,9 @@ namespace ValveResourceFormat.Renderer.Particles.Emitters
                 var emissionMaxValue = emissionMax.NextNumber(particleSystemState);
                 var emissionRate = emissionMinValue + noise * (emissionMaxValue - emissionMinValue);
 
-                // Aggregate emission into num of particles to emit
-                particlesToEmit += Math.Clamp(emissionRate, 0, emissionMaxValue) * frameTime; // Limit the amount of particles to emit at once in case of refocus
+                // Aggregate emission into num of particles to emit.
+                // Min/Max instead of Clamp because a negative m_flOutputMax must not throw.
+                particlesToEmit += Math.Max(0f, Math.Min(emissionRate, emissionMaxValue)) * frameTime; // Limit the amount of particles to emit at once in case of refocus
 
                 // If nr of particles to emit is > 0, emit it
                 while (particlesToEmit > 1.0f)
--- a/Renderer/ParticleRenderer/INumberProvider.cs
+++ b/Renderer/ParticleRenderer/INumberProvider.cs
@@ -211,7 +211,8 @@ namespace ValveResourceFormat.Renderer.Particles
         public PerParticleCountNormalizedNumberProvider(ParticleDefinitionParser parse) { attributeMapping = new AttributeMapping(parse); }
         public float NextNumber(ref Particle particle, ParticleSystemRenderState renderState)
         {
-            return attributeMapping.ApplyMapping(particle.ParticleID) / Math.Max(renderState.ParticleCount, 1);
+            // Mapping input ranges for this provider type are authored in normalized 0-1 space
+            return attributeMapping.ApplyMapping(particle.ParticleID / (float)Math.Max(renderState.ParticleCount, 1));
         }
     }
 
--- a/Renderer/ParticleRenderer/INumberProvider.cs
+++ b/Renderer/ParticleRenderer/INumberProvider.cs
@@ -211,8 +211,9 @@ namespace ValveResourceFormat.Renderer.Particles
         public PerParticleCountNormalizedNumberProvider(ParticleDefinitionParser parse) { attributeMapping = new AttributeMapping(parse); }
         public float NextNumber(ref Particle particle, ParticleSystemRenderState renderState)
         {
-            // Mapping input ranges for this provider type are authored in normalized 0-1 space
-            return attributeMapping.ApplyMapping(particle.ParticleID / (float)Math.Max(renderState.ParticleCount, 1));
+            // Mapping input ranges for this provider type are authored in normalized 0-1 space.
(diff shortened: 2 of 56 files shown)


kristiker:
Needs rebase over #1214 which fixes the same things

kristiker:
I'm only uncertain about RenderTrails changes. Needs to be compared to s2 renderer.

Renderer/ParticleRenderer/Operators/PositionLock.cs
```diff
@@ -46,57 +46,68 @@ public PositionLock(ParticleDefinitionParser parse) : base(parse)
...
             foreach (ref var particle in particles.Current)
             {
-                var transform = transformInput.NextTransform(ref particle, particleSystemState);
-                var transformPosition = Vector3.Multiply(transform.Translation, componentScale.NextVector(ref particle, particleSystemState));
+                var startTime = ParticleCollection.RandomWithExponentBetween(particle.ParticleID, startTimeExp, startTimeMin, startTimeMax);
+                var endTime = ParticleCollection.RandomWithExponentBetween(particle.ParticleID, endTimeExp, endTimeMin, endTimeMax);
 
-                if (previousTransformPosition.X == float.MaxValue)
+                // Fully locked until startTime, fading the lock out until endTime
+                var lockStrength = particle.Age <= startTime
```
w1tcherrr: I think this should be something like:

```
var lifeFraction = MathUtils.Saturate(particle.NormalizedAge);

var lockStrength = lifeFraction <= startTime
    ? 1f
    : 1f - MathUtils.Saturate(MathUtils.Remap(lifeFraction, startTime, endTime));
```

because the startTime and endTime are normalized and not absolute seconds.

Renderer/ParticleRenderer/Operators/Spin.cs
```diff
@@ -51,7 +51,8 @@ public override void Operate(ParticleCollection particles, float frameTime, Part
         {
             foreach (ref var particle in particles.Current)
             {
-                particle.SetScalar(ParticleField.Roll, particle.Rotation.Z + GetSpinRate(particle.Age) * frameTime);
+                // Rotation is stored in radians, the spin rate in degrees per second
+                particle.SetScalar(ParticleField.Roll, particle.Rotation.Z + float.DegreesToRadians(GetSpinRate(particle.Age)) * frameTime);
```
w1tcherrr: For some reason S2 multiplies by another 2 pi here (wtf?). So this spins 2 pi too slow. That means using 57 as spin rate degrees turns around once a second (57 * 2 * pi ≈ 360). 

Particle to test:
[spin_test.zip](https://github.com/user-attachments/files/29713125/spin_test.zip)

Renderer/Utils/MathUtils.cs
```diff
@@ -37,6 +37,22 @@ public static float RemapRange(float x, float inputMin, float inputMax, float ou
...
+        /// <param name="x">Value to remap.</param>
+        /// <param name="inputMin">Input range minimum.</param>
+        /// <param name="inputMax">Input range maximum.</param>
+        /// <param name="outputMin">Output range minimum.</param>
+        /// <param name="outputMax">Output range maximum.</param>
+        /// <returns>Value remapped to the output range, clamped to it.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static float RemapValClamped(float x, float inputMin, float inputMax, float outputMin, float outputMax)
+        {
+            return float.Lerp(outputMin, outputMax, Saturate(Remap(x, inputMin, inputMax)));
```
w1tcherrr, line 59: Should add

```
if (inputMin == inputMax) {     
    return x >= inputMax ? outputMax : outputMin;
}
```

because S2 treats a degenerate range as a threshold, but here it returns inputMax (constant value).

E.g. this particle remaps speed 10->10 to radius 4->40 on a particle with speed 0. PET returns radius 4, this returns 40.

[remap_speed_test.zip](https://github.com/user-attachments/files/29713644/remap_speed_test.zip)

Renderer/ParticleRenderer/INumberProvider.cs
```diff
@@ -211,7 +211,8 @@ class PerParticleCountNormalizedNumberProvider : INumberProvider
         public PerParticleCountNormalizedNumberProvider(ParticleDefinitionParser parse) { attributeMapping = new AttributeMapping(parse); }
         public float NextNumber(ref Particle particle, ParticleSystemRenderState renderState)
         {
-            return attributeMapping.ApplyMapping(particle.ParticleID) / Math.Max(renderState.ParticleCount, 1);
+            // Mapping input ranges for this provider type are authored in normalized 0-1 space
+            return attributeMapping.ApplyMapping(particle.ParticleID / (float)Math.Max(renderState.ParticleCount, 1));
```
w1tcherrr: It should be particle.Index not particle.ParticleID, because the ParticleID is the lifetime spawn counter, not the slot in the (currently) alive list.

E.g. this particle sets radius from "Percent of Total Count" (= the particle's slot in the alive list, 0..1). Since ParticleID keeps increasing on every spawn, it quickly exceeds the alive count and every particle clamps to the top of the range.

[number_normalized_test.zip](https://github.com/user-attachments/files/29714014/number_normalized_test.zip)

Renderer/ParticleRenderer/Initializers/RingWave.cs
```diff
@@ -54,8 +54,8 @@ private float GetNextAngle(int particlesPerOrbit, int maxParticles)
         {
             if (evenDistribution)
             {
-                // -1 falls back to the collection's maximum particle count.
-                var perOrbit = Math.Max(1, particlesPerOrbit == -1 ? maxParticles : particlesPerOrbit);
+                // Unset (-1) or invalid counts fall back to the collection's maximum particle count.
+                var perOrbit = Math.Max(1, particlesPerOrbit <= 0 ? maxParticles : particlesPerOrbit);
```
w1tcherrr: This is wrong, old behaviour is correct. Only -1 is the sentinel value for using max particle count. 

[ringwave_test.zip](https://github.com/user-attachments/files/29714131/ringwave_test.zip)

Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
```diff
@@ -38,6 +38,7 @@ public override void Start(Action particleEmitCallback)
             this.particleEmitCallback = particleEmitCallback;
 
             time = 0f;
+            particlesToEmit = 0f;
```
w1tcherrr: S2 starts this at 1, otherwise the later `while (particlesToEmit > 1.0f)` never emits for a very short / slow emitter (rate * duration <= 1).

Renderer/ParticleRenderer/Emitters/NoiseEmitter.cs
```diff
@@ -68,8 +69,9 @@ public override void Emit(float frameTime, ParticleSystemRenderState particleSys
                 var emissionMaxValue = emissionMax.NextNumber(particleSystemState);
                 var emissionRate = emissionMinValue + noise * (emissionMaxValue - emissionMinValue);
 
-                // Aggregate emission into num of particles to emit
-                particlesToEmit += Math.Clamp(emissionRate, 0, emissionMaxValue) * frameTime; // Limit the amount of particles to emit at once in case of refocus
+                // Aggregate emission into num of particles to emit.
+                // Min/Max instead of Clamp because a negative m_flOutputMax must not throw.
+                particlesToEmit += Math.Max(0f, Math.Min(emissionRate, emissionMaxValue)) * frameTime; // Limit the amount of particles to emit at once in case of refocus
```
w1tcherrr: The upper clamp should not exist actually, because the min/max mapping already targets the correct range. With an inverted range (e.g. max = 0, min = 100) Min(rate, outputMax) kills the emitter. Should just be `Math.Max(0f, emissionRate)`.

[noise_emitter_test.zip](https://github.com/user-attachments/files/29714377/noise_emitter_test.zip)

Renderer/ParticleRenderer/Initializers/PointList.cs
```diff
@@ -69,30 +69,24 @@ private Vector3 GetParticlePosition(ParticleSystemRenderState particleSystem)
...
-                    : pointList.Count;
+                // An open path interpolates across Count - 1 segments; a closed loop has one
+                // extra segment wrapping back to the first point
+                var numPathSegments = closedLoop
+                    ? pointList.Count
+                    : pointList.Count - 1;
 
-                // Percentage of path completed / 100
+                // Percentage of path completed
                 var pathCompletion = relativeParticleNumber / (float)numPointsOnPath;
```
w1tcherrr: Should divide by the numPointsOnPath - 1 (so last point on the path gets a particle).

Also S2 spaces the particle by arc length, not segment index (it sums the real segments lengths, including the wrap segment for closed loops). Then finds the segment containing fraction * totalLEngth, so paths with unequal segments clump in this code.

E.g. 20 particles on a 3-point path with 40 + 360 unit segments - ingame it gets spread evenly over all 400 units, here 10 squeeze into the first 40 units and the line stops 36 units short of the end (due to the first issue).

[pointlist_test.zip](https://github.com/user-attachments/files/29714635/pointlist_test.zip)

w1tcherrr (changes requested):

w1tcherrr:
One more particle to test my commit...

[trail_clamp_test.zip](https://github.com/user-attachments/files/29717602/trail_clamp_test.zip)

PR #1214, opened 2026-07-03
Render path_particle_rope cables and expand the particle system

w1tcherrr:
# Render path_particle_rope cables and expand the particle system

Renders `path_particle_rope` / `path_particle_rope_clientside` cables, which VRF didn't draw at all before. Instead of drawing the flat authored spline, the spline is sampled into a chain of particles (through `ParticleSnapshot`) and run through a real particle simulation: `C_OP_BasicMovement` gravity pulls it down and a new `C_OP_RopeSpringConstraint` holds neighbours near a rest length derived from `slack`. The chain settles into a catenary, then freezes (`m_flStopSimulationAfterTime`, so there is no ongoing cost), and is drawn by the new `C_OP_RenderCables`.

Degenerate data (missing effect, zero radius/spacing, single-node path) is skipped with a warning (and can be found in official content such as `maps/a2_drainage.vmap_c` in HLA). `C_OP_RenderRopes` is aliased onto the tube renderer so HLA rope effects render as tubes instead of nothing.

rope_comparison

cable_comparison

cable_comparison_2

## Particle system work this needed

- **Constraints**: a new particle-function category (`ParticleFunctionConstraint`, backed by `m_Constraints`) plus `C_OP_RopeSpringConstraint`. Constraints run in a work-list loop bounded by `C_OP_BasicMovement`'s `m_nMaxConstraintPasses`, and get their own group in the viewer's operator tree.

- **Force generators as their own class**: `C_OP_AttractToControlPoint`, `C_OP_CurlNoiseForce` and `C_OP_RandomForce` were plain operators that wrote `Velocity`, which `C_OP_BasicMovement` overwrote every frame, so they did not work properly. They now have a base class (`ParticleFunctionForceGenerator`, backed by `m_ForceGenerators`) and run from inside `C_OP_BasicMovement`, accumulating an acceleration into a per-particle `ForceAccumulator` that is integrated as `ForceScale * (gravity + accumulated)` and cleared. A dt/prevDt inertia correction keeps momentum framerate-independent. `ForceScale` of 0 pins a particle, which is how the rope's endpoints hold while the middle droops.

- **Initial velocity works now**: spawn velocity is encoded into the Verlet state as `PositionPrevious = Position - Velocity * dt`. `C_OP_MaxVelocity` also writes its clamp back into the Verlet state now (it was a no-op before).

- **Pre-simulation and freezing**: systems burst-simulate `m_flPreSimulationTime` at spawn in fixed `m_flMaximumTimeStep` substeps (so anything using it settles before first draw), and can freeze after `m_flStopSimulationAfterTime`. `m_flMaximumTimeStep` of 0 no longer freezes the effect (falls back to the default). This is how the ropes are presimulated (the slack and constraint).

- **Snapshot-driven spawning**: `C_INIT_InitFromCPSnapshot` gained `m_nManualSnapshotIndex` / `m_nSnapShotIncrement`, and its sampling was extracted into shared `CPSnapshotSampler` / `CPSnapshotFields`. `C_OP_InstantaneousEmitter` can spawn exactly one particle per snapshot element (`m_nSnapshotControlPoint`, `m_strSnapshotSubset`), which gives one particle per spline sample.

- **Preview control-point drivers**: the preview now applies a system's `m_drivers` (`m_iControlPoint`, `m_iAttachType`, `m_vecOffset`, `m_angOffset`, `m_attachmentName`), so effects whose radius/slack come from a driven control point (like the default cable) preview correctly instead of as an invisible zero-width line. This makes the previews in the particle viewer match the PET better.

- **Control points carry a full rotation**: previously only a forward vector, now a quaternion fed from map-entity angles (and preview `m_angOffset`). This is needed especially for sprites with a fixed orientation.

## Sprite rendering fixes

While lighting the cable (new `particle_cable.vert/frag.slang`, wrapped-diffuse + Blinn-Phong matching `cables.vfx`, with light-probe and sun-shadow support), several `C_OP_RenderSprites` issues were fixed:

- `SCREEN_Z_ALIGNED`, `WORLD_Z_ALIGNED`, `ALIGN_TO_PARTICLE_NORMAL` and `SCREENALIGN_TO_PARTICLE_NORMAL` orientations now orient as named, instead of all four falling back to the plain `SCREEN_ALIGNED` billboard.

- Screen-size clamps: `m_flMinSize` / `m_flMaxSize` were parsed but never used; tiny effects like impact flashes need the minimum to be visible from a reasonable distance.

- Blend and depth-write state is set explicitly by each renderer (`RenderSprites`, `RenderTrails`, `RenderCables`) instead of relying on leftover state.

sprite_orientation_fix

## Smaller fixes

- Operators that scale a particle's spawn value now read the value the initializers set, not the uninitialized default. The initial-state snapshot was never taken after the initializer pass, so `C_OP_FadeOut`/`FadeIn`/`FadeAndKill`, `C_OP_InterpolateRadius`, `C_OP_ColorInterpolate`, `C_OP_LerpScalar`/`LerpVector` and every `PARTICLE_SET_SCALE_INITIAL_VALUE`/`ADD_TO_INITIAL_VALUE` set method scaled from the default alpha/radius/color instead of the authored one. The snapshot is now captured at the end of emission (and the particle's slot index set at spawn), fixing many stock effects across every game.

- Particle map entities feed their full transform into control point 0 (position, forward, full rotation), so effects on rotated entities blow in the authored direction instead of axis-aligned. The scene node stays translation-only (the sim runs in world space, so rotating it would move the bounding box off the particles).

- `C_OP_ContinuousEmitter`'s `m_flEmitRate` is re-evaluated every frame instead of cached at spawn, so control-point/curve-driven rates respond.

- `PF_TYPE_PARTICLE_DETAIL_LEVEL` providers use the highest authored LOD tier (matching `r_particle_max_detail_level`) instead of the lowest.

- `C_INIT_CreateFromParentParticles` inherits the parent's position and velocity (from its Verlet step) instead of only scaling `Velocity`, and walks the parent index (`m_flIncrement`) with the correct fractional step and modulo.

- `C_INIT_RingWave` distributes particles evenly when `m_flParticlesPerOrbit` is unset (the -1 default stacked them all at angle 0), and gained its `m_flInitialSpeedMin`/`m_flInitialSpeedMax` outward push.

- `C_INIT_InitialVelocityNoise` adds to velocity instead of overwriting it (it was deleting earlier initializers' velocity), samples noise at the per-particle coordinate (`m_flNoiseScaleLoc`/`m_vecOffsetLoc`/`m_flOffset` were unparsed, so a whole tick shared one velocity), and remaps signed noise into the authored range.

- Emitters evaluate start time, duration and count/rate against live system state instead of a static default (detail-level and control-point driven fields were frozen before).

- The particle viewer gained a Detail Level selector (Low to Ultra, default Ultra) driving `PF_TYPE_PARTICLE_DETAIL_LEVEL`, which documents why VRF shows more particles than the game's default Medium. The PET defaults to Medium.

- `C_OP_AttractToControlPoint` reads the modern `m_TransformInput` (attachment points, not just `m_nControlPointNumber`), and its default `m_fFalloffPower` is fixed.

- New providers/expressions: `PVEC_TYPE_CP_DELTA`, `VECTOR_EXPRESSION_LERP`, `SCALAR_EXPRESSION_EQUAL`/`GT`/`LT`, `PF_MAP_TYPE_ROUND` (`NEAREST`/`FLOOR`/`CEIL`), and `PF_TYPE_CONTROL_POINT_SPEED`. `SCALAR_EXPRESSION_DIVIDE` by zero outputs 0 instead of inf/NaN.

- `C_INIT_RandomVectorComponent` clamps its component index into [0, 2] (Dota sometimes authors -1, which runs as component 0).

- `GLParticleViewer`'s operator tree lists groups in PET order: pre-emission, emitters, initializers, operators, force generators, constraints, renderers.

## Testing

Added `Tests/PathParticleRopeTest.cs` for the `pathnodes` / `pathnoderadiusscales` / `pathnodecolors` / `pathnodepinsenabled` blob parser (real `cs_italy` data plus edge cases: trailing commas, degenerate/empty blobs, numeric vs word pin flags, misaligned data).

Every `.vpcf_c` across all Source 2 games (~122k files) was swept through load, preview construction and a few simulation ticks with zero exceptions. All particles load in the engine without exceptions.

Reference effects:

These effects (among many others of course) can be used to see the differences this PR makes.

| Effect | Before (master) | After (this branch) |
|---|---|---|
| CS2 `particles/unified_weapon_fx/uweapon_muzzleflash_pist_fire.vpcf` | Nothing renders | Renders decently |
| CS2 `particles/inferno_fx/explosion_incen_ground_splash07a.vpcf` | Fully broken | Renders, but badly |
| CS2 `particles/explosions_fx/explosion_hegrenade_dirt_particulate.vpcf` | Does not render | Renders, but badly |
| CS2 `particles/explosions_fx/explosion_smoke_disperse.vpcf` | Renders, but misaligned | Renders well; alignment and placement correct |
| Dota 2 `particles/units/heroes/hero_jakiro/jakiro_ambient_eye_glow_fire.vpcf` | Force application wrong, so particles disperse incorrectly | Renders well |
| CS2 `particles/entity/path_particle_cable_default.vpcf` | Does not render | Renders well |
| CS2 `particles/inferno_fx/molotov_child_flame01a.vpcf` | Nothing renders | Renders, but badly |
| CS2 `particles/explosions_fx/explosion_hegrenade_embers.vpcf` | Renders, but a bit worse | Renders somewhat well |
| HLA `particles/environment/xen_slime_rope_path.vpcf` | Does not render | Renders well |
| CS2 `particles/explosions_fx/bumpmine_detonate_sparks.vpcf` | Renders, but at wrong scale | Renders very well |
| CS2 `particles/explosions_fx/bumpmine_detonate_ring.vpcf` | Renders, but badly (wrong forces, alpha) | Renders perfectly |
| CS2 `particles/explosions_fx/bumpmine_detonate.vpcf` | Renders, but badly (wrong forces, alpha) | Renders very well |
| HLA `particles/blood_impact/blood_impact_synth_01_dust.vpcf` | Renders, but wrong scale | Renders very well |
| HLA `particles/fire_01/env_embers_tiny.vpcf` | Does not render at all | Renders well (the `m_flEmissionScale` emission scaling is still unimplemented) |
| HLA `particles/jean/flare_smoke.vpcf` | Renders well, but wrong scaling | Renders perfectly |
| HLA `particles/environment/light_glow_01.vpcf` | Renders well | Renders well |

## Notes
- Performance seems fine. Adds around 0.3 seconds on my PC when loading Dust 2, since the pre-settling of the ropes has to be calculated.
- The cable shader is completely vibe-coded. I have no clue how it works and could not figure it out with RenderDoc. It does look decent on most lighting, but also plain wrong in some cases.


Renderer/Renderer/World/WorldLoader.cs
```diff
@@ -947,10 +974,14 @@ void LoadEntity(string classname, Entity entity, Matrix4x4 parentTransform, bool
...
 
+                            // The simulation runs in world space (the entity rotation reaches it through
+                            // control point 0), so the node transform must stay translation-only to keep
+                            // the bounding box wrapping the actual particles.
                             var particleNode = new ParticleSceneNode(scene, particleSystem, particleSnapshot)
                             {
                                 Name = particle,
-                                Transform = Matrix4x4.CreateTranslation(origin),
+                                Transform = Matrix4x4.CreateTranslation(transformationMatrix.Translation),
+                                ControlPointTransform = transformationMatrix,
```
kristiker: This feels like a backwards fix.

> keep the bounding box wrapping the actual particles.

That should managed by the node internally

Renderer/ParticleRenderer/ParticleDefinitionParser.cs
```diff
@@ -35,11 +35,30 @@ public readonly ParticleDefinitionParser[] Array(string k)
...
     }
 
-    private readonly float Float(string k) => Data.GetFloatProperty(k);
+    private readonly float Float(string k)
+    {
+        // Newer content authors many scalar fields as full float-input structures; evaluate those.
+        if (Data.GetSubCollection(k) is { IsCollection: true })
+        {
+            return NumberProvider(k)?.NextNumber() ?? default;
+        }
```
kristiker: We should log them and update their schema. This is only calling number provider once, when it should be called every frame.

PR #1218, opened 2026-07-04
Model export: morphs/flex rules, turn animations, cloth physics

### LINKED ISSUE #653: Tarp mesh decompiling
### Describe the bug

`models\personal\sergiy\cloth_tarp3_mesh` can be used is for cloth simulations in s2 (https://youtu.be/EMny6q6nooo?si=kd0OimvoG9-pSS_R), but decompiling the model, and recompiling it in the editor breaks this functionality.

Here's what the cloth simulation is supposed to look like...
image

And here's what it looks like when decompiled...
image
image

_No response_

Alyx

`models\personal\sergiy\cloth_tarp3_mesh`

### Expected behavior

Expected the mesh to include functionality for cloth simulations

### What version of VRF are you using? On what platform?

6.0.2421

iwasinminedream:
## Overview

This PR improves `ModelExtract` so that decompiled `.vmdl` sources recompile into a much closer match of the original compiled model. It was developed and tested **primarily on Dota 2 models** (heroes and wearable items, e.g. `legion_commander` — turns/morphs/skins/187 sequences, `dark_willow` — cloth sheet, `primal_beast_base` — cloth chains + collision capsules, `dark_carnival_legion_commander` wearable — LODs/activity modifiers/morphs); other Source 2 games were not tested.

**Disclosure:** all changes in this PR were made with the assistance of AI. Validation was done by round-tripping: decompile with VRF → recompile with `resourcecompiler` → compare compiled blocks (DATA/PHYS/ASEQ/ANIM/MRPH) and animation curves via glTF export.

### Morphs (face flexes)
- Morph targets are exported as sparse DMX delta states with a `combinationOperator` and `bindState`, so a recompile rebuilds the MRPH block — previously all face flex data was lost.
- Compiled flex rules (RPN op programs) are decompiled back into the expression strings ModelDoc's `MorphRule` accepts; `MorphControl` nodes are emitted for controllers with non-default ranges, including reconstructed paired controllers (e.g. `eyeDownAndUp` → `eyeUp`/`eyeDown`).

### Animations
- Sequence activity modifiers (`m_activityArray[1..]`, e.g. `injured`, `ether`) are now emitted as `ActivityModifier` child nodes; previously they were dropped entirely.
- 1D blend sequences are reconstructed as `AnimTurn` (lookFrame pattern) or generic `1DBlend` nodes instead of being flattened into baked `AnimFile`s — turn leaning works again on recompiled models.
- Local-hierarchy overrides (`m_hierarchyArray`) are parsed and emitted as `LocalHierarchy` children (weapon detaching in death animations).
- Animation DMX file names are prefixed with the model name so sibling models in one directory no longer overwrite each other's `idle`/`run`/`attack` DMX files.

### Skins and LODs
- Material groups (`m_materialGroups`) are decompiled into `DefaultMaterialGroup` + `MaterialGroup` nodes with `BaseMaterialRemap` entries — hero skins/personas survive a recompile.
- Meshes present in several LOD levels are repeated per `LODGroup` instead of being collected into a synthetic `LODGroupAll` node (which is not a real ModelDoc node type).

### Cloth physics (initial work)
- New `FeModel` parser for the compiled soft-body data (`m_pFeModel`) in the PHYS aggregate.
- When the cloth surface is recoverable, it is exported as a cloth proxy-mesh DMX (per-vertex cloth paints, chain skinning) referenced by a `ClothProxyMeshFile` with back-solved joints, plus a `Softbody` with recovered `ClothParams` and collision capsules/spheres.
- Otherwise bone-chain cloth is emitted as version-2 `ClothChain` nodes with per-joint goal strength/damping/gravity, collision radii and stray radii recovered from the compiled node integrators, plus a generated (disabled) sheet grid as an authoring aid.

## Known limitations
- **Cloth is not fully accurate yet**: for some models the recompiled cloth behaves like the original, for others it behaves incorrectly. This is the beginning of physics decompilation work, not the end.
- Models with embedded physics that cannot be reconstructed yet (e.g. ragdoll-only aggregates) get a small placeholder physics shape so the recompiled model still allocates a PHYS block and the CTRL `embedded_physics` reference.
- Everything was verified on Dota 2 content only.


dirtkiller23:
idk about removing LODGroupAll as it's a real modeldoc node, it's used for non LOD meshes that are needed to display on all LOD levels/regardless of LOD level. also you should join the source2 engine discord server

kristiker:
Can you add resourcecompiler tests please. The Test project should detect local Source 2 Workshop Tools and validate sucessful decompile.

iwasinminedream:
You're right, thanks — `LODGroupAll` is indeed a real node (`CModelDocLODGroupAll`, "A group of meshes that are present in all lod levels" in modeldoc_editor.dll). I dropped that change from the PR and restored the upstream behaviour.

iwasinminedream:
Added `ResourceCompilerTest` — it detects local Source 2 Workshop Tools installations through Steam (currently Dota 2 and CS2), decompiles a set of models from the game VPKs (covering morphs, activity modifiers, turn/1D blend sequences, material groups, LODs and cloth), stages them into a temporary addon, compiles them with `resourcecompiler`, and validates the round-trip: the compile must succeed, the output must load back as a valid model resource, and flex controllers, material groups, sequences and cloth (FeModel) may not be lost. When no Workshop Tools installation is found the tests are skipped, so CI is unaffected. All 7 cases pass locally against both Dota 2 and CS2.

Writing the CS2 cases caught a real bug: agent models (e.g. `ctm_sas`) lost their flex controllers on recompile — their morphs carry no texture atlas (the flex channels only drive DotToMorph bone constraints), and rule/controller parsing was gated behind the atlas load. Fixed by parsing flex rules/controllers directly from the MRPH data.

One known gap: SDF-only cloth colliders (e.g. the ak47's `m_SDFRigids` FeModel) are not reconstructed yet.

iwasinminedream:
Decompile cloth physics (FeModel) into editable ModelDoc source

Decompiles the compiled cloth (FeModel in the PHYS block) back into editable ModelDoc source (ClothChain / ClothProxyMeshFile / ClothNode / ClothSpring + proxy mesh), so a decompiled model recompiles into working cloth and opens in ModelDoc.

The physics decompile isn't perfect yet, and I've run out of ideas on where to dig next.

- Simple models usually reconstruct fully and round-trip exactly (some are byte-exact and covered by tests).
- Complex models reconstruct well but not fully — a few nodes come out off by one or two. Everything still recompiles with 0 errors.

I'm not planning to keep pushing the physics for now, because I hit problems I don't know how to solve:

- Some cloth data is per-face paint applied at compile time (cloth_make_rods, cloth_bend_stiffness) and simply isn't in the compiled model — the exact tri/quad split can't be recovered.
- A couple of per-rod fields (flWeight0, flRelaxationFactor) exist in the compiled data but aren't valid authoring inputs — the compiler ignores whatever you set.
- The compiler adds/prunes a few nodes on its own that I can't reproduce from authoring (the ±1–2 above).
- Models compiled with an older compiler reference old parameters, and I don't know how to turn an old parameter into its modern equivalent on recompile. Example: some older models ship flAnimationVertexAttraction = 5–15 (the old solver path), but the current compiler clamps goal_strength to [0,1] and can't produce any value above 1 — I found no authoring input that maps the old value back. Those models reconstruct correctly by value but feel slightly stiffer.

Everything else — topology, rods, integrators, twists, fit matrices, skin weights, bone-chain cloth — round-trips fine.

I've attached a schema of everything I reverse-engineered about the solver and compiler, in case someone wants to pick this up later.
[CLOTH_SOLVER_SCHEMA.md](https://github.com/user-attachments/files/29755546/CLOTH_SOLVER_SCHEMA.md)
and yes much batch of its stuff generated with ai

w1tcherrr:
Cloth decompile is a big undertaking. Closing this PR since I have a newer work in progress branch.

PR #1220, opened 2026-07-06
Panorama writing/construction

Kampfkarren:
Allows for creating Panorama objects with arbitrary data. Not really sure if the constructor API for it is best.


kristiker (approved):

ValveResourceFormat/Resource/ResourceTypes/Panorama.cs
```diff
@@ -46,6 +46,19 @@ public class NameEntry
         /// <inheritdoc/>
         public override BlockType Type => BlockType.DATA;
 
+        /// <inheritdoc />
+        public Panorama() { }
+
+        /// <summary>
+        /// Initializes a new instance.
+        /// </summary>
```
kristiker, line 57: Could use a better xmldoc though

PR #1223, opened 2026-07-08
Adaptive shadow map resolution for dynamic lights

dr3fty:


kristiker:
Lights are no longer dropped due to failed shadow map slot, instead only if max light count is reached? Great change if so.

kristiker:
Seems like not, but since the shadow readjusts now the light will fail to allocate a region less often.

PR #1227, opened 2026-07-10
CS2 bomb blast visualization

krotter0:
White and yellow arrows only, couldn't figure out how phase+bomb power corresponds to damage or radius.

https://github.com/user-attachments/assets/b71b42eb-e58f-48be-b4bc-b540069e6750


xPaw:
Do we actually want it to fade? I know it simulates the delay in game, but it might be nicer to just be able to see how far it's reaching without waiting for it to pulse.

krotter0:
We could have a minimum alpha threshold like this, a bit of fading helps with the visibility of the blast's path I think

https://github.com/user-attachments/assets/7f893ea3-8299-4222-ad97-a5047848964e

Renderer/Renderer/SceneNodes/CS2BombDamageSceneNode.cs
```diff
@@ -0,0 +1,291 @@
...
+    /// <param name="bombsiteIndex">Index of the bombsite. Index 0 is not guaranteed to be bombsite A.</param>
+    public CS2BombDamageSceneNode(Scene scene, BombDamageData bombDamageData, int bombsiteIndex) : base(scene)
+    {
+        creationTime = DateTime.Now;
+
+        shader = Scene.RendererContext.ShaderLoader.LoadShader("vrf.cs2_baked_bomb_damage");
+
+        GL.CreateVertexArrays(1, out vaoHandle);
+
+        var buffers = ArrayPool<int>.Shared.Rent(2);
```
xPaw: stackalloc, also arraypool can return sizes higher than requested (so the gl call below will do many more allocs)

Renderer/Renderer/SceneNodes/CS2BombDamageSceneNode.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    private static int RGBAColor(byte r, byte g, byte b, byte a)
+    {
+        return a << 24 | b << 16 | g << 8 | r << 0;
+    }
+
+    /// <inheritdoc/>
+    public override void Render(Scene.RenderContext context)
```
xPaw, line 216: needs a renderpass check i think

Renderer/Renderer/SceneNodes/CS2BombDamageSceneNode.cs
krotter0: GL.CreateBuffers does not accept Spans, but I've fixed the params

kristiker:
ArrayPool uses in init/destruct aren't necessary, I'm removing them.

Especially `ArrayPool<VertexFormat>.Shared`, this leaves a large array pinned in ram when its only used by the bomb scene node initialization (2 uses).

Renderer/Renderer/SceneNodes/CS2BombDamageSceneNode.cs
```diff
@@ -0,0 +1,267 @@
...
+    private void Initialize(BombDamage bombDamageData, int bombsiteIndex)
+    {
+        var vertices = new List<VertexFormat>(bombDamageData.Positions.Length * 4);
+        var indices = new List<int>(bombDamageData.Positions.Length * 6);
+
+        for (var i = 0; i < bombDamageData.Positions.Length; i++)
+        {
+            if (vertices.Count == 0)
+            {
+                boundsMin = boundsMax = bombDamageData.Positions[i];
```
kristiker: Worth checking if the bounds from resource itself is the same, so we dont need to recalculate from positions.

PR #1228, opened 2026-07-10
Fix local addons not being loaded as workshop dependencies

Arenovas:
Previously when loading in the Summit Pavilion Remix maps from SteamVR Home, they'd fail to load their assets due to the first entry being for the included steamvr_home addon. The program would try and fail to find that in the external workshop folder which would also lead to not loading the other addon that is within the external workshop folder.
So I added a check for the steamvr_home string and direct it to load from the proper location.
I definitely feel like this is a dirty hack that could be cleaned and simplified but though I'd at least open this PR and see what else I could do to improve it.

Before the fix:
image

After the fix:
image

### DIFF
--- a/ValveResourceFormat/IO/GameFileLoader.cs
+++ b/ValveResourceFormat/IO/GameFileLoader.cs
@@ -54,7 +54,7 @@ namespace ValveResourceFormat.IO
         private bool ShaderPackagesScanned;
         private bool AttemptToLoadWorkshopDependencies;
         private bool StoredSurfacePropertyStringTokens;
-        private string SteamVRDir;
+        private string? SteamVRDir;
 
         /// <summary>
         /// Gets or sets the current package being processed.
--- a/ValveResourceFormat/IO/GameFileLoader.cs
+++ b/ValveResourceFormat/IO/GameFileLoader.cs
@@ -54,6 +54,7 @@ namespace ValveResourceFormat.IO
         private bool ShaderPackagesScanned;
         private bool AttemptToLoadWorkshopDependencies;
         private bool StoredSurfacePropertyStringTokens;
+        private string SteamVRDir;
 
         /// <summary>
         /// Gets or sets the current package being processed.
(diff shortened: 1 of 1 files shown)


ValveResourceFormat/IO/GameFileLoader.cs
```diff
@@ -189,12 +190,20 @@ public virtual (string? PathOnDisk, Package? Package, PackageEntry? PackageEntry
...
+                                }
+                                else
                                 {
-                                    AddPackageToSearch(dependencyVpkPath);
+                                    var dependencyId = (uint)dependency;
+                                    var dependencyVpkPath = Path.Join(workshopRoot, $"{dependencyId}", $"{dependencyId}.vpk");
+                                    if (File.Exists(dependencyVpkPath))
+                                    {
+                                        AddPackageToSearch(dependencyVpkPath);
+                                    }
```
kristiker, line 204: Is the dependency list just numbers and strings. Numbers being workshop addons and strings being local addons?

Arenovas: Yep pretty much. And I think the only string that's ever used as a dependency is steamvr_home.

ValveResourceFormat/IO/GameFileLoader.cs
```diff
@@ -189,12 +190,20 @@ public virtual (string? PathOnDisk, Package? Package, PackageEntry? PackageEntry
                         {
                             foreach (var (_, dependency) in dependencies)
                             {
-                                var dependencyId = (uint)dependency;
-                                var dependencyVpkPath = Path.Join(workshopRoot, $"{dependencyId}", $"{dependencyId}.vpk");
-
-                                if (File.Exists(dependencyVpkPath))
+                                uint addon = 0;
+                                bool workshopAddon = uint.TryParse((string?)dependency, out addon);
```
xPaw: I think you can check the type of kvvalue here

Arenovas: Yeah I was wondering if there was a built in way to do that but decided to figure something out myself first.
If there is a method to do so I'd be happy to implement that in!

Issue #257, opened 2020-07-12
Render particles referenced by models


xPaw:
Dropship model in HL:A has this in DATA:
```
		m_keyValueText = "<!-- kv3 encoding:text:version{e21c7f3c-8a33-41c5-9977-a76d3a32aa0d} format:vmdlkeys3:version{141d1e66-eebf-4440-be28-a74cd93426eb} -->
{
	particles_list = 
	[
		{
			name = resource:\"particles/creatures/dropship/dropship_thruster.vpcf\"
			attachment_point = \"RearThruster_L\"
			attachment_type = \"point_follow\"
			attachment_offset = [ 0.0, 0.0, 0.0 ]
		},
		{
			name = resource:\"particles/creatures/dropship/dropship_thruster.vpcf\"
			attachment_point = \"RearThruster_R\"
			attachment_type = \"point_follow\"
			attachment_offset = [ 0.0, 0.0, 0.0 ]
		},
		{
			name = resource:\"particles/creatures/dropship/dropship_thruster.vpcf\"
			attachment_point = \"RearThruster_R_Lower\"
			attachment_type = \"point_follow\"
			attachment_offset = [ 0.0, 0.0, 0.0 ]
		},
		{
			name = resource:\"particles/creatures/dropship/dropship_thruster.vpcf\"
			attachment_point = \"RearThruster_L_Lower\"
			attachment_type = \"point_follow\"
			attachment_offset = [ 0.0, 0.0, 0.0 ]
		},
	]
}"
```

PR #1233, opened 2026-07-15
Render particles referenced by models and allow parenting to bones

w1tcherrr:
Implements #257

Renders the particle systems listed in a model's `particles_list`.

Also fixes the entity `parentname` path so it now falls back to a bone when no attachment of that name exists.

Implements only the 4 particle attachment types actually used by any shipped particles.

See e.g.
HLA -> `models/creatures/combine_dropship/dropship.vmdl_c`
Deadlock -> `models/abilities/familiar/familiar_helpinghands_aura_statue.vmdl_c` and `models/npc/shrine_sapphire/shrine_sapphire.vmdl_c`

### DIFF
--- a/Renderer/Renderer/World/WorldNodeLoader.cs
+++ b/Renderer/Renderer/World/WorldNodeLoader.cs
@@ -106,12 +106,6 @@ namespace ValveResourceFormat.Renderer.World
                     };
 
                     scene.Add(modelNode, false);
-
-                    foreach (var modelParticleNode in ParticleSceneNode.CreateModelParticles(scene, model, modelNode))
-                    {
-                        modelParticleNode.LayerName = "Particles";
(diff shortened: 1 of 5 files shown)


Renderer/Renderer/World/WorldNodeLoader.cs
```diff
@@ -106,6 +106,12 @@ public void Load(Scene scene)
                     };
 
                     scene.Add(modelNode, false);
+
+                    foreach (var modelParticleNode in ParticleSceneNode.CreateModelParticles(scene, model, modelNode))
```
xPaw: I can't imagine world node models having particles

w1tcherrr: That's true, I removed it

PR #1236, opened 2026-07-18
Support additive animations and NM clip root motion, retarget AG2 clips in glTF, and export morph weight animation

### LINKED ISSUE #1046: Option to correctly preview and export additive animations

Image

I’m noticing that animations like sh_rifl_stand_firing in cs2 make the character model curl up in the viewer. My guess is that this animation is additive, and in-game it’s applied only to a specific set of bones (rather than the full skeleton).

Is there a way to add a “Fix additive animation” option in both the viewer and the export workflow, so the animations preview/export correctly—i.e., matching how they look in-game?

If not, is it possible to get some expert tips on how I can use these additive animations in for example unreal engine

_No response_

w1tcherrr:
# Support additive animations and NM clip root motion, retarget AG2 clips in glTF, and export morph weight animation

On master, selecting an additive (delta) animation collapses the model into a ball at the origin, because every animation is played as an absolute pose. 

This PR detects additive animations for both ag1/ag2 and composes them over the bind pose in the viewer and in glTF exports.

The PR also implements `.vnmclip` root motion.

Lastly, it unifies legacy animgraph and animgraph2 clips behind one `Animation` abstract base class.

## Additive animations

- animgraph1 does not have a direct additive flag. It lives in the animation graph, so the new `AnimationGraph1Additive` walks the graph (`CAddAnimNode` additive children, additive `CAimMatrixAnimNode`) and `Model.GetAllAnimations` stamps `Animation.IsAdditive` from it. Both graph layouts are handled: the uncompiled editor layout HLA ships and the compiled layout Deadlock ships (`CAddUpdateNode`/`CAimMatrixUpdateNode`, `m_hSequence` handles resolved against the model's sequence table). Subtract-node subtrees are excluded: they derive their delta at runtime from absolute inputs, so the sequences below them are not delta content.

- animgraph2 clips have a `m_bIsAdditive` flag. It is also used to mark same-named animgraph1 sequences. This covers deadlock, since their animgraph2 models have no animgraph1 graph to walk, but embed ag1 duplicates of their clips as retarget sources (`ag1/retarget_from/...`), compiled from the same source animation and referenced by nothing. Their own sequence data does not carry this information, so the same-named clip's `m_bIsAdditive` is the only place their additive-ness is set, and they inherit it through the shared name.

- Composition is centralized in `Animation.ComposeAdditiveOverBindPose`. The viewer sets the "Additive (over bind pose)" checkbox per animation and keeps it as a manual override. The info labels now always show the `Additive:` and `Root Motion:`.

## NM clip root motion

- `AnimationClip` parses `m_rootMotion` into per-frame position and yaw. glTF exports bake it into the root bones. NM root motion natively carries vertical travel, so Z is kept there (the legacy movement system stays planar).

- Clip framerate is fixed: FPS is `(NumFrames - 1) / duration`, exactly 30.0 on every shipped Deadlock and CS2 clip. Master's `NumFrames / duration` was slightly wrong.

- NM float curves are decoded into named per-frame channels.

## AG2 clip playback and glTF retargeting

- AG2 clips animate a separate NM skeleton and are remapped onto the model skeleton by bone name. That rule now lives in one place (`SkeletonRetargeter`, with `FramePose` for world-pose accumulation), shared by the renderer and the glTF exporter. On master the exporter copies the clip skeleton's local transforms onto the model joints, so every exported AG2 clip comes out broken. Clips are now retargeted by world pose, matching the renderer.

- `AnimationController` now routes playback through a single runner chosen per animation.

## Morph weight animation in glTF

On master, glTF exports write morph targets but never animate them, so facial animations look static. A new pass evaluates each mesh's flex rules per frame against the decoded flex controller values and writes the weights as morph channels. 

## Architecture

- `Animation` is now an abstract base with `SequenceAnimation` (AG1) and `ClipAnimation` (AG2). Accessors (`RequiresRetarget`, `TargetSkeletonName`, `SupportsMixerAdditive`) keep consumers generic, and `Model.GetAllAnimations` returns graph clips too, so there is one collection for everything a model can play.

- Most of the reusable animation code moves into the core library: the `Animation` classes, `BaseAnimationController`, the world-pose walk (`FramePose`) and the bone-name retargeter (`SkeletonRetargeter`). The renderer and both exporters share one implementation of the pose and retarget math, and library consumers get additive composition, root motion and clip retargeting without referencing the renderer package.

## Comparisons

Master on the left, this branch on the right.

### Additive animations

HLA `models/characters/gman/gman.vmdl_c`, animation `facial_noise`: collapsed vs composed over bind:

additive_delta

Deadlock `models/heroes_wip/frank/frank.vmdl_c`, animation `flinch_melee_left_add`: composed over bind instead of collapsing:

additive_delta_2

Deadlock `models/heroes_staging/bebop/bebop.vmdl_c`, animation `flinch_melee_left_add`: additive inherited from the same-named graph clip; also shows the always-visible `Additive:` / `Root Motion:` info lines:

flinch_bugged

### NM clip root motion

Deadlock `models/heroes_wip/frank/clips/dash_ground.vnmclip_c`: root motion moves the skeleton and defaults on. Also visible: master plays the clip at 30.58 fps, this branch at 30:

root_motion_checkbox

Deadlock `models/heroes_staging/bebop/clips/turn_left.vnmclip_c`: the 90 degree root yaw is applied:

root_motion_yaw

Deadlock `models/heroes_wip/frank/clips/flinch_melee_left_add.vnmclip_c`: 32 fps on master vs the correct 30:

fps_fix

### Skeleton overlay

HLA `models/characters/gman/gman.vmdl_c`, animation `gman_turn_exit`: root motion moves the skeleton overlay with the model instead of leaving it at the origin:

root_motion_bones

HLA `models/characters/gman/gman.vmdl_c`, animation `gman_trailer`: cloth bones follow the body instead of freezing at the origin:

cloth_bones

### glTF export

Deadlock `models/heroes_staging/bebop/bebop.vmdl_c`, clip `turn_left`, exported to glb: retargeted by world pose instead of coming out mangled:

bebop_turn

Deadlock `models/heroes_wip/frank/frank.vmdl_c`, clip `flinch_melee_left_add`, exported to glb: composed over bind pose instead of collapsing:

frank_flinch

## Notes

- Graph-unreferenced orphan sequences (dead content such as `gman_head_aim_matrix`, referenced by no graph in either layout) still need the manual checkbox. The modern engines have a per-clip `m_bAnimGraphAdditive` field for exactly this, which is honored, but Valve sets it on no shipped content.
- NM clip secondary animations (weapon overlays on foreign skeletons) are not written into model glTF exports, same as master.
- Hitboxes do not follow root motion (they derive from the pose, not the node transform).


kristiker (commented):
This is really hard to reivew. I think the AnimationController Runners changes are welcome. The way they are nested with subcontrollers is terrible. This probably needs to be merged first and in another PR if possible.

w1tcherrr:
Split into: #1249, #1253 and #1260.

PR #1242, opened 2026-07-20
Sound event player / 3D Audio system

### LINKED ISSUE #879: Add soundevent support in model viewer when playing animation

There's event data on every animation, some of them related to playing sounds at specific keyframes. We already parse these, here is how they look:
```
{
    m_nFrame = 19
    m_flCycle = 0.019895
    m_EventData = 
    {
        name = "ScriptedSeq.Gman_Void_Step_Walk"
        attachment = "snd_foot_l"
        stop_on_seq_change = false
        use_layer_sequence = false
    }
    m_sOptions = ""
    m_sEventName = "AE_CL_PLAYSOUND_ATTACHMENT"
},
```

We can already play vsnd files. What we need is parsing the soundevent files, and building a simple system that plays vsnd files given the soundevent name (`ScriptedSeq.Gman_Void_Step_Walk`). Then on every ModelSceneNode update we check if a new animation event has occured, and play the sound if the event is of type `AE_CL_PLAYSOUND...`.

Sound plays when the model animates.

_No response_

kristiker:
Introduced a new audio subsystem to the application, enabling sound playback on 3D scenes.

- Added audio playback support on `ValveResourceFormat.Renderer`. The audio device is provided by client, such as our viewer, which currently implements `NAudioDevice`, a WASAPI-based audio output device using NAudio. This can be later changed to OpenAL or SDL audio for cross compatibility.
- Most of the sound event work is originally from @krotter0

**Audio Mixer and Playback System:**
- `AudioMixer` manages and mixes active sound events into a single stereo stream, supporting spatialization and real-time updates from the render/game thread.
- `CachedSound` stores fully decoded, interleaved float samples, supporting loop points for seamless playback.
- `AudioConverter` converts decoded audio to the mixer's output format, handling channel count and sample rate conversion with linear interpolation.

#### TODO
- [x] Play map `point_soundevent`
- [x] Play map `env_soundscape`
- [x] Support sound event types from multiple games

#### Sound On
https://github.com/user-attachments/assets/357444e6-083a-47bf-b9c0-e5b829abf092


ValveResourceFormat/Resource/ResourceTypes/ModelAnimation2/AnimationClip.cs
```diff
@@ -128,6 +134,14 @@ private void ReadClip(KVObject clipData)
                 };
             }
 
+            var events = clipData.GetArray("m_events") ?? [];
+            Events = new NmClipEvent[events.Count];
+
+            for (var eventIndex = 0; eventIndex < events.Count; eventIndex++)
+            {
+                Events[eventIndex] = NmClipEvent.Build(events[eventIndex], Duration);
+            }
```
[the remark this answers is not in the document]
kristiker: not in this pr

Renderer/Audio/SoundEventPlayer.cs
```diff
@@ -0,0 +1,911 @@
...
+
+        try
+        {
+            using var wrapped = new MemoryStream();
+            wrapped.Write("\"__root\"\n{\n"u8);
+            stream.CopyTo(wrapped);
+            wrapped.Write("\n}\n"u8);
+            wrapped.Position = 0;
+
+            return KeyValues1Serializer.Deserialize(wrapped);
```
kristiker, line 388: does vkv really not support multiple root kvs

edit: yes https://github.com/ValveResourceFormat/ValveKeyValue/issues/134

Renderer/Audio/Decoders/MpegFrameWalker.cs
```diff
@@ -0,0 +1,291 @@
+using NLayer;
+
+namespace ValveResourceFormat.Renderer.Audio.Decoders;
+
+/// <summary>
+/// A reusable <see cref="IMpegFrame"/> that walks MP3 frames directly over a byte buffer, replacing
+/// NLayer's stream-based <c>MpegFile</c>/<c>MpegStreamReader</c> front end.
+/// </summary>
+internal sealed class MpegFrameWalker : IMpegFrame
```
kristiker, line 15: Why the MP3 path doesn't use NLayer's MpegFile

NLayer still does all the actual decoding — MpegFrameDecoder (huffman, IMDCT, synthesis).

What **is** replaced is only the stream front end (MpegFile/MpegStreamReader), by MpegFrameWalker, which implements NLayer's public IMpegFrame and walks frames directly over the vsnd's bytes.

The reason is per-decode allocation. MpegFile constructs a reader and a decoder per file and can't be rebound to another stream, and the reader also allocates per frame: a 4-byte sync buffer in FindNextFrame, a MpegFrame object, and a SaveBuffer() copy of each frame's bytes on seekable streams (where the frame chain is additionally retained for the whole file). Measured over 50 × 200-frame files that's ~200 KB per file. The walker plus a [ThreadStatic] MpegFrameDecoder is ~0 at steady state, since the two decode threads are long-lived and reuse both.

This matters even though decoding happens off the render thread — GC is process-wide, so decode-thread garbage stalls rendering too. Also we can't guarantee that the loader precaches all sounds, so some might still stream mid rendering.

Tradeoff: NLayer's front end also handles iTunSMPB gapless from ID3v2, RIFF-wrapped MPEG, free-format frames, and resyncs more strictly. The walker reads Xing/LAME gapless from the first frame and rejects free-format, which is fine for compiled vsnd payloads.

PR #1245, opened 2026-07-21
Add KV3 version 5 and generalized KV3 compression serialization

theokyr:
## What changed

- Add public binary KV3 version and compression settings.
- Preserve version 5 and its source compression method when rewriting loaded blocks.
- Serialize KV3 version 5 with its two-buffer layout and uncompressed, raw LZ4, or Zstandard output.
- Encode binary blobs as uncompressed data, independent Zstandard frames, or chained 16 KiB LZ4 frames as required by the selected compression method.
- Keep newly constructed and pre-version-5 KV3 blocks on the existing version 4 uncompressed behavior.
- Reject unsupported version and compression combinations instead of silently changing the requested output.

## Why

Current Deadlock resources use binary KV3 version 5 in multiple resource blocks and with mixed compression methods. VRF can read these blocks, but its writer previously emitted every block as uncompressed version 4. This prevented a read/write cycle from preserving the source format and made it impossible to create genuine version 5 output.

## Usage

New `BinaryKV3` blocks remain version 4 and uncompressed by default. Version 5 output can be selected explicitly:

```csharp
var block = new BinaryKV3(root, KV3IDLookup.Get("generic"))
{
    SerializationVersion = KV3BinaryVersion.Version5,
    SerializationCompressionMethod = KV3BinaryCompressionMethod.Zstd,
};

block.Serialize(output);
```

Blocks loaded from version 5 input automatically retain their source version and compression method when serialized again.

## Validation

- Round-trip tests cover uncompressed, LZ4, and Zstandard version 5 output.
- Synthetic coverage includes scalar values, flags, nested collections and arrays, non-object roots, empty values, and a binary blob larger than 16 KiB to exercise chained LZ4 frames.
- The existing Zstandard fixture verifies source-setting preservation.
- A 1,338-byte fixture from Deadlock build `24302421` verifies that its `RED2`, `DATA`, and `FLCI` blocks remain version 5 with their original compression methods and unchanged semantic KV3 output.
- The Deadlock fixture is a compiler/schema resource with no Steam ID, match ID, username, gameplay history, or other player-specific data.
- `dotnet format` completed for the changed projects.
- The full solution builds with zero warnings and errors.
- The complete test suite passes with 315 tests passed and one existing fixture-dependent test skipped.
- The built Source2Viewer CLI successfully reads freshly generated version 5 LZ4 output containing a blob larger than 16 KiB.

Fixture SHA-256: `59dd48f7be3b55c105b59f52325a93b7ffec077f07a1942e314907d20b27d64c`.

## Disclosure

Most of this PR was built by `gpt-5.6-sol medium` through the Codex harness.


theokyr:
Keeping this as draft until I can verify Deadlock doesn't crash with various modified KV3 V5's. Currently investigating programatically generating hero builds.

theokyr:
image
Programatically generated builds now possible!

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.Serialization.cs
```diff
@@ -141,19 +178,207 @@ public override void Serialize(Stream stream)
...
+
+            if (SerializationCompressionMethod == KV3BinaryCompressionMethod.Uncompressed)
+            {
+                return input;
+            }
+
+            if (SerializationCompressionMethod == KV3BinaryCompressionMethod.Zstd)
+            {
+                return CompressZstd(input);
+            }
```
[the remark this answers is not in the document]
theokyr: Reject unless a Valve-produced v5/Zstd file containing external blobs demonstrates that layout.

theokyr:
So far so good with my testing replacing Deadlock KV3 V5s. Requesting review @kristiker @xPaw

xPaw:
v4 also supports same compression methods, so ideally this code should be more generalized.

theokyr:
Generalized compression serialization for v4 and v5. v4 uncompressed is still the default. Now supports serializing aggregate chained LZ4 blobs, or combined main/blob Zstd.

Regarding Copilot reviews:
- No Valve provided files supported the case for per-blob frames.
- Applied the checked arithmetic/comment fixes.

Codex ran 406 CS2, Dota 2, and Deadlock KV4 and KV5 blocks through a round-trip test fixture, by rewriting and rereading each, and comparing binary headers with the pristine versions, under all six version/compression combinations. Full test suite: 325 passed, 1 skipped.

Requesting re-review @xPaw

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.cs
```diff
@@ -196,6 +236,18 @@ private void ReadBuffer(int version, BinaryReader reader)
             var format = KV3IDLookup.GetByValue(new Guid(reader.ReadBytes(16)));
 
             var compressionMethod = reader.ReadUInt32();
+
+            if (version == 5)
+            {
+                SerializationVersion = KV3BinaryVersion.Version5;
+                SerializationCompressionMethod = compressionMethod switch
```
xPaw: i think existing compression method needs to be assigned into this enum for all versions, so that remaining reader code would also use the enum.

and im not entirely a fan of KV3BinaryVersion, but not sure what to do about it yet.

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.Serialization.cs
```diff
@@ -33,11 +38,14 @@ public SerializationContext()
                 Bytes4Writer = new BinaryWriter(Bytes4, System.Text.Encoding.UTF8, leaveOpen: true);
                 Bytes8Writer = new BinaryWriter(Bytes8, System.Text.Encoding.UTF8, leaveOpen: true);
                 TypesWriter = new BinaryWriter(Types, System.Text.Encoding.UTF8, leaveOpen: true);
+                ObjectLengthsWriter = new BinaryWriter(ObjectLengths, System.Text.Encoding.UTF8, leaveOpen: true);
                 BinaryBlobsWriter = new BinaryWriter(BinaryBlobs, System.Text.Encoding.UTF8, leaveOpen: true);
             }
 
             public int GetStringId(string str)
             {
+                CountStringIds++;
```
xPaw: Shoudnt this use StringMap.Count? incrementing it on every get call seems wrong.

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.cs
theokyr: Moved this around a bit and changed the defaults to preserve compression for other versions too [`ff1cd35`](https://github.com/ValveResourceFormat/ValveResourceFormat/pull/1245/commits/ff1cd35e3b27b628d4dae7de9d82633ffd4a1294). Is your issue the existence of a version enum, or its naming scheme?

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.Serialization.cs
theokyr: Fixed in [`ff1cd35`](https://github.com/ValveResourceFormat/ValveResourceFormat/pull/1245/commits/ff1cd35e3b27b628d4dae7de9d82633ffd4a1294)

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.cs
xPaw: Yeah let's remove the version enum for now.

And can you fix the tests to test both 4 and 5 versions? I think that should work without duplicating the code, and tests both paths.

theokyr: Fixed in [`d889882`](https://github.com/ValveResourceFormat/ValveResourceFormat/pull/1245/commits/d8898821bb0d6181f66570e5a4d06ae58041700c)

theokyr:
These checks seemed to have failed due to GitHub CI 503's

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.Serialization.cs
```diff
@@ -78,80 +85,318 @@ public override void Serialize(Stream stream)
...
+            writer.Write(blockCompressedSizes.Count * sizeof(ushort));
+            writer.Write(buffer1.Length);
+            writer.Write(compressed ? compressedBuffer1.Length : 0);
+            writer.Write(buffer2.Length);
+            writer.Write(compressed ? compressedBuffer2.Length : 0);
+            writer.Write((int)context.Bytes1.Length);
+            writer.Write((int)context.Bytes2.Length / 2);
+            writer.Write((int)context.Bytes4.Length / 4 - 1);
+            writer.Write((int)context.Bytes8.Length / 8);
+            writer.Write(context.StringMap.Count);
```
xPaw, line 234: is this actually correct? reader has it as unknown value.

this also means the test `Assert.That(BitConverter.ToInt32(data, 104), Is.EqualTo(5));` is pretty silly, we shouldn't be testing like that.

Tests/KeyValuesTest.cs
```diff
@@ -48,6 +49,444 @@ public void TestBinaryKV3_Serialization()
...
+                Resource = null!,
+                SerializationVersion = 5,
+                SerializationCompressionMethod = KV3BinaryCompressionMethod.Lz4,
+            };
+
+            using var stream = new MemoryStream();
+            binaryKV3.Serialize(stream);
+            var data = stream.ToArray();
+
+            Assert.That(BitConverter.ToInt32(data, 68), Is.EqualTo(4 * sizeof(ushort)));
```
xPaw, line 249: Remove these asserts that try to test ints at specific offsets.

ValveResourceFormat/Resource/ResourceTypes/BinaryKV3.Serialization.cs
```diff
@@ -78,80 +85,318 @@ public override void Serialize(Stream stream)
...
+            var compressedBuffer2 = CompressMainBuffer(buffer2);
+            var compressed = SerializationCompressionMethod != KV3BinaryCompressionMethod.Uncompressed;
+            var countObjects = checked((ushort)(context.ObjectLengths.Length / sizeof(int)));
+            var countArrays = checked((ushort)context.CountArrays);
+            var stringBytesLength = 0;
+
+            foreach (var str in context.Strings)
             {
-                context.BinaryBlobs.WriteTo(writer.BaseStream);
+                stringBytesLength = checked(stringBytesLength + System.Text.Encoding.UTF8.GetByteCount(str) + 1);
```
xPaw, line 205: Make BuildVersion5Buffer1 return stringBytesLength, because it already has the length.

PR #1247, opened 2026-07-26
F_USE_NEW_BLENDING implementation

w1tcherrr:
Basic implementation of F_USE_NEW_BLENDING (from csgo_environment_blend). Looks a bit scuffed, since its missing a lot of parameters, but the general blend math should match closely.

20718C~1

cache_uvonly

### DIFF
--- a/Renderer/Shaders/csgo_environment_features.slang
+++ b/Renderer/Shaders/csgo_environment_features.slang
@@ -12,6 +12,7 @@
 
     #define F_SHARED_COLOR_OVERLAY 0 // csgo_environment_blend
     #define F_ENABLE_LAYER_3 0 // csgo_environment_blend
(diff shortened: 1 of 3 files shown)


Renderer/Shaders/csgo_environment_features.slang
```diff
@@ -12,6 +12,7 @@
 
     #define F_SHARED_COLOR_OVERLAY 0 // csgo_environment_blend
     #define F_ENABLE_LAYER_3 0 // csgo_environment_blend
+    #define F_USE_NEW_BLENDING 0 // csgo_environment_blend
```
kristiker: I would use a dynamic branch. I'm not sure why valve switched back.

w1tcherrr: Done

kristiker (approved):

PR #1249, opened 2026-07-26
(Anim Part 1):  Play animations through one player per skeleton instead of nesting animation controllers

w1tcherrr:
Part one of splitting #1236, as @kristiker requested. This is only the playback rework. I removed the nesting instead of wrapping it, so the runner abstraction is gone.

## Architecture

On master, an `AnimationController` playing an animgraph2 clip delegates to another full `AnimationController` held in `SubController.Handler`. Playback state has two owners and is copied between them every frame.

In this PR `AnimationPlayer` is a new class owning clip time, the mixer, the frame cache and one skeleton's pose. It now never owns another player, so playback state lives in exactly one object. `AnimationController` owns one player per skeleton its model can be animated from, picks the right one in `SetAnimation`, and forwards to it.

`BaseAnimationController` is deleted. Since the nesting is gone there is no second controller type, so its pose and bind pose members fold into `AnimationController`. The world-pose walk moves into the library as `FramePose.ComputeWorldSubtree`.

This does not add the runner abstraction from #1236. Removing the nesting leaves one class and a single dispatch point, which is what the runners were meant to provide.

The frametime multiplier was squared on the external path. The parent scaled `timeStep`, then the nested controller scaled it again by the copy of the multiplier it had been handed. Now this is fixed.

The nested controller's `ApplyInverseKinematics` no longer runs. It was inert: both IK guards early-return unless `TwistConstraints` or `EnableFirstPersonLegs` are set, and neither is ever set on an external skeleton. The controller still applies IK to the remapped pose.

## Next PRs

- **Part two:** split `Animation` into `SequenceAnimation` (AG1) and `ClipAnimation` (AG2). Will be Library only without behaviour changes.
- **Part three:** the rest of #1236, additive animations, NM clip root motion, AG2 retargeting in glTF, morph weight export.


Tests/Renderer/AnimationControllerTest.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                Looping = false,
+            };
+
+            var walk = model.GetEmbeddedAnimations().First(a => a.Name == "box_creature_leggy_walk");
+            controller.SetAnimation(walk);
+
+            // The clip runs 25 frames at 30fps, so a full second overshoots its end and pauses it.
+            Assert.That(controller.Update(1f), Is.True);
+            Assert.That(controller.IsPaused, Is.False, "the mixer only observes the finished clip on the next tick");
```
kristiker: Sounds like weird behavior. Maybe wasn't touched by this PR, but intuitively the controller should report that it is paused, after the previous update just exhausted its clip time.

kristiker:
There are no bugs as far as I can tell. Previous code wasn't the best so this should be okay.

Might be worth looking into S2 schema for inspiration: https://s2v.app/SchemaExplorer/cs2/client/AnimationAlgorithm_t
https://s2v.app/SchemaExplorer/cs2/client/CBaseAnimGraphController

Tests/Renderer/AnimationControllerTest.cs
kristiker: Fixed

kristiker (approved):

PR #1258, opened 2026-07-31
Add --game arg for CLI

Egezenn:
Allows proper resource resolution for files that are outside of the respective game's directories.

### DIFF
--- a/CLI/Decompiler.cs
+++ b/CLI/Decompiler.cs
@@ -53,6 +53,7 @@ namespace CLI
         private TextureCodec TextureDecodeFlags;
         private string[] FileFilter = [];
         private bool ListResources;
+        private string? GameInfoPath;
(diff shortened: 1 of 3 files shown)


kristiker:
What about `--game` like the resourcecompiler tool. User can pass gameinfo.gi both should work.

xPaw:
Can you give some examples when you're actually exporting files outside of the game for this to he useful?

Egezenn:
> What about --game like the resourcecompiler tool. User can pass gameinfo.gi both should work.

Will change it

> Can you give some examples when you're actually exporting files outside of the game for this to he useful?

If someone sent you a compiled asset (for whatever reason) you'd have to move the file inside a game directory for it to decompile with full references via Source2Viewer-CLI. Niche case.

PR #1260, opened 2026-07-31
(Anim Part 3): Additive animations, NM clip root motion, glTF clip retargeting and morph weight export

### LINKED ISSUE #1046: Option to correctly preview and export additive animations

w1tcherrr:
Part three of splitting #1236.

## Additive animations

Additive animations play as absolute poses, so selecting one collapses the model into a ball at the origin.

AG2 clips carry `m_bIsAdditive`. AG1 sequences carry nothing: additive is a property of the graph slot, so the graph is walked for sequences feeding add nodes and additive aim matrices. Both layouts are handled, the uncompiled one HLA ships and the compiled one Deadlock ships. Subtract subtrees are skipped, their inputs are absolute. AG1 sequences sharing a name with an additive clip inherit the flag, which is how Deadlock models get theirs.

Composition is shared by the renderer and both exporters. The viewer seeds a checkbox per animation and keeps it overridable.

## NM clips

- **Root motion** is parsed (position and yaw). The viewer moves the model and skeleton overlay by it, checkbox defaults on, glTF bakes it into the root bones including vertical travel.
- **Frame rate** is (frames - 1) over duration. Shipped Deadlock and CS2 clips land on exactly 30 instead of 30.58, and decompiled event times follow from it.
- **Float curves** are decoded into named per-frame channels.

## glTF export

- AG2 clips are retargeted by world pose and bone name instead of having local transforms copied across, which mangled them.
- Secondary skeletons (the weapon in a viewmodel clip) get their own joints instead of being dropped.
- Morph targets are animated: flex rules are evaluated per frame and written as morph weights. They were static before.

## NM clip decompile

Secondary skeleton tracks were dropped. They are now written into the same DMX, where the compiler expects them.

## Other fixes

- Flex weights and movement no longer leak across an animation switch.
- Cloth bones follow the body instead of freezing at the origin.
- The viewer's time readout uses the animation's duration.

## Where the animations come from

Models now return their graph clips alongside their embedded animations, so the viewer and both exporters work off one list instead of each loading clips themselves. Map previews skip them, so opening a map no longer loads every model's graph and clips.

## Comparisons

Master on the left, this branch on the right.

### Additive animations

HLA `models/characters/gman/gman.vmdl_c`, animation `facial_noise`: collapsed vs composed over bind:

additive_delta

Deadlock `models/heroes_wip/frank/frank.vmdl_c`, animation `flinch_melee_left_add`: composed over bind instead of collapsing:

additive_delta_2

### NM clip root motion

Deadlock `models/heroes_wip/frank/clips/dash_ground.vnmclip_c`: root motion moves the skeleton and defaults on. Also visible: master plays the clip at 30.58 fps, this branch at 30:

root_motion_checkbox

Deadlock `models/heroes_staging/bebop/clips/turn_left.vnmclip_c`: the 90 degree root yaw is applied:

root_motion_yaw

Deadlock `models/heroes_wip/frank/clips/flinch_melee_left_add.vnmclip_c`: 32 fps on master vs the correct 30:

fps_fix

### Skeleton overlay

HLA `models/characters/gman/gman.vmdl_c`, animation `gman_turn_exit`: root motion moves the skeleton overlay with the model instead of leaving it at the origin:

root_motion_bones

HLA `models/characters/gman/gman.vmdl_c`, animation `gman_trailer`: cloth bones follow the body instead of freezing at the origin:

cloth_bones

### glTF export

Deadlock `models/heroes_staging/bebop/bebop.vmdl_c`, clip `turn_left`, exported to glb: retargeted by world pose instead of coming out mangled:

bebop_turn

Deadlock `models/heroes_wip/frank/frank.vmdl_c`, clip `flinch_melee_left_add`, exported to glb: composed over bind pose instead of collapsing:

frank_flinch

## Notes

- Every `.vnmclip` in CS2 and Deadlock parses without errors, about 12k files. Viewer, glTF export and decompile were checked on HLA, Deadlock and CS2 models.
- Dead sequences that no graph references, such as `gman_head_aim_matrix`, cannot be identified as additive and still need the checkbox. Newer branches have a per-clip flag for this, which is honored, but Valve sets it on almost nothing that shipped.
- Hitboxes do not follow root motion.


kristiker:
If this changes additive animations to export as blended over bindpose, I don't think that is necessary. We can add hints that the anim contains additive transforms, but otherwise its up to the consumer of the glb to blend it like our viewer.

Or at least it should be optional.

w1tcherrr:
> If this changes additive animations to export as blended over bindpose, I don't think that is necessary. We can add hints that the anim contains additive transforms, but otherwise its up to the consumer of the glb to blend it like our viewer.
> 
> Or at least it should be optional.

I changed it to default to off. Additive animations export as the delta only, with some extras on the glTF animation flagging it as that.

`"extras": { "additive": true, "additive_base": "bindpose", "additive_composed": false }`

New CLI option --gltf_compose_additive that turns the composition on.

For the UI I think it should be added once the new UI is done in the future and some custom glTF settings dialogue is added (could also contain stuff like whether to export animations, etc...).

PR #1285, opened 2026-08-09
Fix reference grid translucency

w1tcherrr:
Grid draws bugged over translucent stuff, especially noticeable for particles. Depth mask change due to Z-fighting of Grid on particles aswell.

Old:
bugged
New:
correct


kristiker (commented):
We want to fix this using OIT algorithms. Besides that grid should be a scene node for easier sorting.

Issue #1351, opened 2026-09-15
Decompiling s&box shader dont work anymore


PolEpie:
### Describe the bug

I tried decompile a sbox shader, but since they moved their compiling shader process to slang internally, we cant decompile it anymore

### Expected behavior

Decompile like before

### Relevant log output or exception with stacktrace

```shell
[14:00:08.163] [MainForm] Opening C:\Program Files (x86)\Steam\steamapps\common\sbox\core\shaders\colorgrading.shader_c
[14:00:08.166] [Console] Added folder "C:\Users\polep\Documents\s&box projects\nrp_map" to game search paths
[14:00:14.088] [ExtractProgressForm] Failed to extract 'colorgrading.shader_c': System.InvalidOperationException: Shader extract cannot continue without at least a features file.
   at ValveResourceFormat.IO.ShaderExtract..ctor(ShaderCollection shaderCollection) in /_/ValveResourceFormat/IO/ShaderExtract.cs:line 131
   at ValveResourceFormat.IO.FileExtract.Extract(Resource resource, IFileLoader fileLoader, IProgress`1 progress) in /_/ValveResourceFormat/IO/FileExtract.cs:line 261
   at GUI.Forms.ExtractProgressForm.ExtractFile(Resource resource, String inFilePath, String outFilePath, Boolean flatSubfiles) in /_/GUI/Forms/ExtractProgressForm.cs:line 499


[14:00:32.649] [RendererContext] Initializing GPU texture decoder...
[14:00:33.508] [RendererContext] GPU: NVIDIA GeForce RTX 4080/PCIe/SSE2, Driver: 4.6.0 NVIDIA 610.47, OS: Microsoft Windows NT 10.0.26200.0
[14:00:49.293] [MainForm] Opening C:\Program Files (x86)\Steam\steamapps\common\sbox\core\shaders\complex.shader_c
[14:00:49.294] [Console] Added folder "C:\Users\polep\Documents\s&box projects\nrp_map" to game search paths
[14:00:55.661] [ExtractProgressForm] Failed to extract 'complex.shader_c': System.InvalidOperationException: Shader extract cannot continue without at least a features file.
   at ValveResourceFormat.IO.ShaderExtract..ctor(ShaderCollection shaderCollection) in /_/ValveResourceFormat/IO/ShaderExtract.cs:line 131
[...]

PR #1352, opened 2026-09-15
Added sbox vfx 66 version support + upgrade feature

PolEpie:
Fix for #1351 

S&box bumped its embedded VCS version from 65 to 66, which `VfxProgramData.VfxCreateFromVcs`
did not handle (it asserted 65 in Debug and silently misparsed in Release). The resulting
exception was swallowed by `SboxShader.Read`, leaving an empty `ShaderCollection`, so
exporting any current s&box shader failed with the misleading
`Shader extract cannot continue without at least a features file`.

Version 66 is identical to 65 except that `VfxCombo` gains two int32 fields between
`FeatureIndex` and the state name count: a feature upgrade flag and the value auto-assigned
to new materials. The 64-byte alias slot now carries the feature's UI category, which the
existing extract code already emits as the third `Feature(...)` argument.

### Changes

- `VfxProgramData`: accept s&box 65 and 66, expose the raw value as `SboxVersion`, and keep
  remapping `VcsVersion` to 64 so all downstream version gates stay correct. Unknown s&box
  versions now throw `UnexpectedMagicException` instead of misparsing.
- `VfxCombo`: read the two new fields into `IsFeatureUpgrade` / `FeatureUpgradeValue`.
- `SboxShader.Read`: dispose and rethrow instead of swallowing parse failures, so future
  format bumps fail at open time with a clear error.
- `ShaderExtract`: emit `FeatureUpgrade( F_NAME, value );` after the matching `Feature`.
- Docs: note VCS 65 and 66 support in the format support guide.

### Testing

- New `sbox66_fur.shader_c` fixture covering v66 combos, UI categories, `FeatureUpgrade`,
  rules and static combo blobs, plus a `ShaderTest.SboxShaderVersion66` test asserting the
  parsed values and the reconstructed `.shader` output.
- The existing v65 fixture is kept, so both versions stay covered.
- Decompiling `complex.shader_c` now produces a full `FEATURES` block instead of throwing


ValveResourceFormat/CompiledShader/VfxProgramData.cs
```diff
@@ -67,6 +67,12 @@ public class VfxProgramData : IDisposable
         /// </summary>
         public int VcsVersion { get; private set; }
 
+        /// <summary>
+        /// Gets the s&amp;box file version, or 0 for Valve files. s&amp;box version 65 matches Valve
+        /// version 64; version 66 adds feature upgrade fields to combos.
```
kristiker: 65 does not match valve 64

PolEpie: You're right, fixed. s&box has its own version lineage (their 65 doesn't carry Valve 65's LayerId fields, and the header has the extra ABI int32), so the VcsVersion = 64 is just normalization for the parse gates, not an equivalence.     
Reworded the comment to say that

PR #1355, opened 2026-09-17
feat: mouselook

Poggicek:


GUI/Types/GLViewers/GLBaseControl.cs
```diff
@@ -389,6 +389,7 @@ public void OnLostFocus(object? sender, EventArgs e)
         Keys.D4 => TrackedKeys.Slot4,
         Keys.E => TrackedKeys.E,
         Keys.F => TrackedKeys.F,
+        Keys.R => TrackedKeys.R,
```
Poggicek: @xPaw @kristiker hammer has this bind as "Z", do we nuke Q/Z Up/down binds in favor of this

kristiker: Keep Z as Down command when a mouse button is held?

PR #1358, opened 2026-09-18
Include CollisionGroup in CLI export node extras

itrisa:
Adding the CollisionGroup to the export of glTF model to allow my and other downstream apps to select geometry based on the collision group.

Also updated the test to check that node.Extras has the expected fields populated.


ValveResourceFormat/IO/Gltf/GltfModelExporter.Physics.cs
```diff
@@ -105,6 +105,7 @@ private void LoadPhysicsMeshes(ModelRoot exportedModel, Scene scene, PhysAggrega
                     node.Extras = new System.Text.Json.Nodes.JsonObject
                     {
                         ["SurfaceProperty"] = surfaceProperty,
+                        ["CollisionGroup"] = collisionAttributes[collisionAttrIndex].GetStringProperty("m_CollisionGroupString"),
```
kristiker, line 108: I think some really old physics have only hash, this will be null.

